PR #60, opened 2025-08-07
Chore: migrate backstage-plugin to new frontend system

jhfgloria:
### Description

This PR includes a new export (`/alpha`) with the new frontend system. This allows not only to keep using the most up to date recommendations from the Backstage team, but also allows integrating these plugin in Portal. The change could be seen as a breaking change but we're exporting the changes in a side-module to allow existing users to keep using the old system as they use to.

In order to adopt the NFS in Backstage users have to start using the new `createApp` helper exported by `@backstage/frontend-defaults` instead of `@backstage/app-defaults`. If not, no changes are needed.

```javascript
import PagerDutyPlugin from '@pagerduty/backstage-plugin/alpha';
...
import { createApp } from '@backstage/frontend-defaults';

const app = createApp({
  features: [
    ApiExplorerPlugin,
    ApiDocsPlugin,
    CatalogPlugin,
    ...
    PagerDutyPlugin,
  ],
});

export default app.createRoot();
```

### Affected plugin

- [x] backstage-plugin

### Type of change

### Checklist

- [x] Changes are documented
- [x] Changes generate *no new warnings*

## Acknowledgement


packages/app/src/index.tsx
```diff
@@ -1,6 +1,13 @@
...
-import App from './App';
 import '@backstage/canon/css/styles.css';
 
-ReactDOM.createRoot(document.getElementById('root')!).render(<App />);
+// Uncomment the lines below if you want to use the old version of the app
+// import App from './App';
+
+// ReactDOM.createRoot(document.getElementById('root')!).render(<App />);
+
+// Uncomment the lines below if you want to use the alpha version of the app
```
sandornagy517, line 10: nit: Comment

jhfgloria: I'll leave it here for future travellers. If you want to flip it between the old and new versions you don't need to know how to do it from memory.

sandornagy517 (approved):

PR #72, opened 2025-09-04
feat: Allow default Backstage auth policy on plugin route

jkueh:
### Description

This is a recreation of https://github.com/PagerDuty/backstage-plugin-backend/pull/114.

Corresponding documentation update: https://github.com/PagerDuty/backstage-plugin-docs/pull/14

Currently, any user that is able to access a Backstage instance that has the PagerDuty plugin installed can make proxied calls to the PagerDuty API without needing a user session in Backstage, e.g.

curl "https://${BACKSTAGE_HOST}/api/pagerduty/services"
The above will return a list of all the services in the account, and presumably can make other proxied API calls to the rest of the PagerDuty API, subject to the permissions of the API token available in Backstage.

This PR adds opt-in support to disable the unauthenticated policy, which [Backstage's own documentation](https://backstage.io/docs/backend-system/core-services/http-router/#using-the-service:~:text=This%20dangerously%20allows%20unauthenticated%20access%20on%20a%20specific%20route) refers to as dangerous.

Ideally this would be more of an opt-out feature, but I don't have visibility into current usage of this plugin, so I'm leaving it as opt-in for now.

I've tested this in a local instance of Backstage, and can confirm I get an HTTP 401 when attempting to run the same curl command as above without a valid Authorization: Bearer ... header.

### Affected plugin

- [ ] backstage-plugin

### Type of change

- [X] Fix (non-breaking change which fixes an issue)

### Checklist

- [X] Changes are documented (https://github.com/PagerDuty/backstage-plugin-docs/pull/14)
- [ ] Changes generate *no new warnings*
- [ ] PR title follows [conventional commit semantics](https://www.conventionalcommits.org/en/v1.0.0/)

## Acknowledgement

### DIFF
--- a/plugins/backstage-plugin-backend/src/plugin.ts
+++ b/plugins/backstage-plugin-backend/src/plugin.ts
@@ -12,7 +12,7 @@ class CatalogFetchApi {
   constructor(
     private readonly logger: LoggerService,
     private readonly auth: AuthService,
-  ) {}
+  ) { }
 
   async fetch(
     input: unknown,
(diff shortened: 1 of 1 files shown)


jhfgloria:
Hello @jkueh! Thanks for flagging this! I think we should flip the enablement logic, making it unaccessible by default and having a property to make it accessible without token. That would make the plugin more secure by default and less secure on demand. In that case we would release a new minor version to flag the change. Could I ask you 2 things too? Since we changed the plugins into a monorepo, new contributions internal and external should be backed by a release note, which can be easily generated by running `yarn changeset`. Could you create a minor version for `backstage-plugin-backend` in this pull request? Also, could you change the base branch from `main` to `next`? Thank you again @jkueh!

jkueh:
Hi @jhfgloria!

Not a problem, will do those shortly.

My initial thinking was that if we're doing a minor release, we probably shouldn't be breaking any existing implementations that rely on access to this plugin being unauthenticated.

That being said, happy to switch it around as part of a minor release if this is more bug-fix territory :)

jkueh:
Awesome, thanks @jhfgloria!

PR #107, opened 2025-12-10
feat: Swap mapping direction

sandornagy517:
### Description

To improve loading performance of the mappings page we swap the direction of the data loaded. Until now were were mapping Backstage entities to PagerDuty services, now we swap this and will do the opposite. In this PR I've updated the mappings table to the new Backstage UI table and updated the way we load the data for the component.

image

### Affected plugin

- [X] backstage-plugin

### Type of change

### Checklist

- [X] Changes are documented

## Acknowledgement


plugins/backstage-plugin-backend/src/apis/pagerduty.ts
```diff
@@ -746,6 +746,20 @@ export async function getServiceByIntegrationKey(
...
 }
 
+export async function getServicesByIds(
+  ids: string[],
+): Promise<PagerDutyService[]> {
+  let services: PagerDutyService[] = [];
+  await Promise.all(
+    Object.entries(EndpointConfig).map(async ([account, _]) => {
+      services = await Promise.all(
+        ids.map(async id => await getServiceById(id, account)),
```
jhfgloria: I think we should probably use this API instead, because we can do it in one request? https://github.com/PagerDuty/service-search-service/blob/master/apps/api/doc/api/service_directory.md#api-service_directory-list_services

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -117,34 +119,38 @@ export async function buildEntityMappingsResponse(
...
-    const entityRef = componentEntitiesDict[service.id]?.ref;
-    const entityName = componentEntitiesDict[service.id]?.name;
-
-    // Check if the service is mapped to an entity in the database
+  componentEntities.items.forEach(ent => {
     const entityMapping = entityMappings.find(
-      mapping => mapping.serviceId === service.id,
+      e =>
+        e.integrationKey ===
+        ent?.metadata.annotations?.['pagerduty.com/integration-key'],
```
jhfgloria: Are there any guarantees that at this point the `integration-key` was set? 🤔 Why not searching by `integration-key` OR `service-id`.

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/StatusCell.tsx
```diff
@@ -0,0 +1,55 @@
...
+    case 'InSync':
+      return 'green';
+    case 'OutOfSync':
+      return 'red';
+    case 'NotMapped':
+      return 'orange';
+    default:
+      return 'gray';
+  }
+}
```
jhfgloria: In JavaScript I always preferred the dictionary approach to the switch one.


```suggestion
const statusDictionary = {'InSync': 'In Sync', ... 'NotMapped': 'Not Mapped'} as const;
const colourDictionary = {'InSync': 'green', ..., 'NotMapped': 'orange'} as const;

function getStatusName(status?: string) {
  return statusDictionary[status] || 'Refresh to Update';
}

function getColorFromStatus(status?: string) {
  return colourDictionary[status] || 'gray';
}
```

jhfgloria (commented):
Left some comments. The PR is quite big (and visual), so probably I would need to test it locally to make more sense of it.

plugins/backstage-plugin-backend/src/apis/pagerduty.ts
```diff
@@ -671,6 +671,76 @@ export async function getServiceById(
...
+      throw new HttpError(
+        'Failed to get service. Caller did not supply credentials or did not provide the correct credentials.',
+        401,
+      );
+    case 403:
+      throw new HttpError(
+        'Failed to get service. Caller is not authorized to view the requested resource.',
+        403,
+      );
+    case 404:
```
jhfgloria, line 725: What happens if at least one of the provided IDs does not exist?

plugins/backstage-plugin-backend/src/apis/pagerduty.ts
```diff
@@ -671,6 +671,76 @@ export async function getServiceById(
   }
 }
 
+export async function getServicesByIdsServiceDirectory(
```
jhfgloria: I don't think we need to expose the calling service from PagerDuty. `getServicesByIdsAndAccount` would be more readable and not expose the fact that the call is made to `ServicesDirectory`.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -684,6 +686,186 @@ export async function createRouter(
...
+            'pagerduty.com/integration-key':
+              entity?.metadata?.annotations?.[
+                'pagerduty.com/integration-key'
+              ] ?? '',
+            'pagerduty.com/service-id':
+              entity?.metadata?.annotations?.['pagerduty.com/service-id'] ?? '',
+          };
+
+          const formattedEntity = {
+            name: entity?.metadata?.name,
```
jhfgloria: Why would the entity be null? Doesn't it come from the `catalog` API? Also, if it is null would be better practice to early return here, instead of dragging the `?` around forever?

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -684,6 +686,186 @@ export async function createRouter(
...
+          // Try to find a service by service ID or integration key
+          let service = null;
+          if (annotations['pagerduty.com/service-id']) {
+            const serviceId = annotations['pagerduty.com/service-id'];
+            const foundServices = pagerDutyServices.filter(
+              s => s.id === serviceId,
+            );
+            if (foundServices.length > 0) {
+              service = foundServices[0];
+            }
```
jhfgloria:
```suggestion
            service = pagerDutyServices.find(s => s.id === serviceId);
```

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -684,6 +686,186 @@ export async function createRouter(
...
+            const integrationKey = annotations['pagerduty.com/integration-key'];
+            const account =
+              entity?.metadata?.annotations?.['pagerduty.com/account'] || '';
+            try {
+              service = await getServiceByIntegrationKey(
+                integrationKey,
+                account,
+              );
+            } catch (e) {
+              // Service might not exist, just continue
```
jhfgloria: I think this type of error is only valid if the error is a 404. Otherwise you will not know if there is a map or not and could decide to map something else.

plugins/backstage-plugin-backend/src/apis/pagerduty.ts
sandornagy517: it is ignored, the rest of the services are returned only

jhfgloria (changes requested):
LGTM! Let's just make some tests in the ServiceMappingComponent (I think it is the topmost component) and test the router.ts new endpoint (which I also think is the topmost backend piece of software). There is no tests in the PR and I think there should be.

plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -56,33 +57,61 @@ async function createDatabase(): Promise<PagerDutyBackendStore> {
 
 describe('createRouter', () => {
   let app: express.Express;
+  let store: PagerDutyBackendStore;
+  const mockCatalogApi = {
```
jhfgloria: Do we need to mock the entire class? 🤔

plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -2273,5 +2302,302 @@ describe('createRouter', () => {
...
+        );
+
+        const response = await request(app)
+          .post('/mapping/entities')
+          .send({ offset: 0, limit: 10 });
+
+        expect(response.status).toEqual(200);
+        expect(response.body).toHaveProperty('entities');
+        expect(response.body).toHaveProperty('totalCount');
+        expect(Array.isArray(response.body.entities)).toBe(true);
```
jhfgloria: Well, `[ ]` is also an array, that doesn't make this test more accurate. I think we should check if the responde contains the expected mapping, and not just check if it is an array. Otherwise I can change the implementation of the controller to just return an empty array and it will pass anyway.

plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -2273,5 +2302,302 @@ describe('createRouter', () => {
...
+          }),
+        );
+
+        const response = await request(app)
+          .post('/mapping/entities')
+          .send({ offset: 20, limit: 5 });
+
+        expect(response.status).toEqual(200);
+        expect(response.body).toHaveProperty('entities');
+        expect(response.body).toHaveProperty('totalCount');
```
jhfgloria: Again, we're not asserting the return of the request, and we should.

plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -2273,5 +2302,302 @@ describe('createRouter', () => {
...
+            offset: 0,
+            limit: 10,
+            search: 'test-component',
+            searchFields: ['metadata.name', 'spec.owner'],
+          });
+
+        expect(response.status).toEqual(200);
+        expect(response.body).toHaveProperty('entities');
+        expect(response.body).toHaveProperty('totalCount');
+      });
```
jhfgloria: Ergh... these tests look all the same thing. The fact that we mock everything in the `mockCatalogApi` doesn't allow for testing much. The fact that you never assert the contents of `entities` is the prove itself that the test doesn't really matter. Tbh, unless we change these tests to actually alter the catalog that is queried, I'd even prefer to just wipe these tests. They're not testing anything in my opinion.

plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -2273,5 +2302,302 @@ describe('createRouter', () => {
...
+
+        const response = await request(app).post('/mapping/entities').send({
+          offset: 0,
+          limit: 10,
+          search: '   ',
+        });
+
+        expect(response.status).toEqual(200);
+        expect(response.body).toHaveProperty('entities');
+        expect(response.body).toHaveProperty('totalCount');
```
jhfgloria: Again, not testing anything.

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+      'Search for components or teams',
+    );
+
+    await act(async () => {
+      fireEvent.change(searchInput, { target: { value: 'my-component' } });
+    });
+
+    await act(async () => {
+      jest.advanceTimersByTime(500);
+    });
```
jhfgloria: I don't think modern RTL tests still require `act` wrapping. `fireEvent` is already wrapped in a `act`. For `advanceTimersByTime` you may be need, because it is not a RTL event.

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+          <MappingsTable />
+        </QueryClientProvider>
+      </ApiProvider>,
+    );
+
+    expect(screen.getByText('In Sync')).toBeInTheDocument();
+    expect(screen.getByText('Out of Sync')).toBeInTheDocument();
+    expect(screen.getByText('Not Mapped')).toBeInTheDocument();
+    expect(
+      screen.getByText('Error occured while fetching service'),
```
jhfgloria, line 179: Does this fit the table column? 🤔

jhfgloria (commented):
Left some comments :)

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.test.tsx
sandornagy517: it does not but gets dotted out, on hover the full text is displayed

plugins/backstage-plugin-backend/src/service/router.test.ts
sandornagy517: I mocked it for TS reasons only

jhfgloria (approved):
LGTM 🚀

PR #120, opened 2026-01-13
feat(backend): add auto-matching algorithm for service mapping

fanfernandes:
## Description

This PR implements an intelligent auto-matching algorithm to map PagerDuty services to Backstage components. The solution uses fuzzy name matching (Jaro-Winkler algorithm) combined with team name and acronym bonuses to provide accurate, confidence-scored mapping suggestions.

**Key Features:**
- Two-level normalization strategy (basic + advanced preprocessing)
- Jaro-Winkler fuzzy string matching for robust name comparison
- Team name matching bonus (+10%)
- Acronym matching bonus (+5%)
- Configurable confidence threshold
- REST API endpoint: `POST /mapping/entity/auto-match`
- Test UI component for validation

**Universal Design Approach:**
The algorithm preserves all meaningful words and only removes structural markers ([prefix], (parentheses)), making it work universally for clients worldwide regardless of language or naming conventions.

**Note:** This implementation was developed collaboratively with Claude Code (Sonnet 4.5) using test-driven development, comprehensive documentation, and iterative refinement. Also this test follow this implementation https://github.com/PagerDuty/backstage-mapping in python already tested.

Closes #DEVECO-522

---

## Which plugins are affected by this PR?

- [x] `@pagerduty/backstage-plugin-backend`
- [x] `@pagerduty/backstage-plugin`
- [ ] `@pagerduty/backstage-plugin-common`

---

## What type of change is this?

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Chore (non-breaking change that does not add functionality)

---

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have tested both in light and dark theme (if applicable)
- [x] I have updated the documentation accordingly (if applicable)
- [x] I have not introduced any warnings in the code or the console
- [x] I have followed the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) guidelines (feat, fix, docs, chore, etc.)

---

### Backend Implementation

**New Files:**
- \`src/utils/normalization.ts\` - Core normalization utilities (274 lines)
- \`src/utils/normalization.test.ts\` - Normalization tests (55 tests)
- \`src/services/dataLoader.ts\` - Data loading from PagerDuty and Backstage (203 lines)
- \`src/services/dataLoader.test.ts\` - Data loader tests (18 tests)
- \`src/services/matchingEngine.ts\` - Matching algorithm with scoring (314 lines)
- \`src/services/matchingEngine.test.ts\` - Matching engine tests (28 tests)

**Modified Files:**
- \`src/service/router.ts\` - Added \`POST /mapping/entity/auto-match\` endpoint (+117 lines)
- \`package.json\` - Added \`@skyra/jaro-winkler\` dependency

**Test Coverage:** 101 unit tests, all passing ✅

### Frontend Implementation

**New Files:**
- \`src/components/PagerDutyPage/AutoMatchTestComponent.tsx\` - Test UI component

**Modified Files:**
- \`src/components/PagerDutyPage/index.tsx\` - Added "Auto-Match Test" tab

**Note:** The AutoMatchTestComponent was created specifically for testing and validation purposes in staging and local environments. It provides a UI to manually trigger the auto-match endpoint and visualize results with confidence scores, match bonuses, and statistics.

### API Endpoint

POST /api/pagerduty/mapping/entity/auto-match

```
Request Body:
{
  "threshold": 80,     // Optional, default 80 (0-100)
  "bestOnly": false    // Optional, default false
}

Response:
{
  "matches": [
    {
      "pdServiceId": "...",
      "pdServiceName": "...",
      "bsComponentRef": "...",
      "bsComponentName": "...",
      "score": 95.5,
      "confidence": "high",
      "exactMatch": false,
      "teamMatch": true,
      "acronymMatch": false
    }
  ],
  "statistics": {
    "totalPdServices": 12,
    "totalBsComponents": 15,
    "totalMatches": 10,
    "exactMatches": 2,
    "teamMatches": 5,
    "acronymMatches": 3
  }
}
```

### Algorithm Details

**Normalization:**
1. Basic: Lowercase, trim, normalize whitespace
2. Advanced: Remove structural markers ([prefix], (parentheses)), preserve meaningful words

**Scoring:**
1. Start with Jaro-Winkler similarity (0-100%)
2. Add +10% bonus if team names match
3. Add +5% bonus if acronyms match
4. Cap final score at 100%
5. Filter by threshold (default: 80%)

**Why Jaro-Winkler?**
- Handles typos and variations well
- Favors matches at the beginning of strings (common in service names)
- Proven algorithm used by PagerDuty's internal tooling

---

## Testing

The feature has been thoroughly tested with both unit tests and manual testing in staging environment.

### Unit Tests
- 101 tests across 3 test suites, all passing ✅
- Comprehensive edge case coverage
- Tests for normalization, data loading, and matching algorithm

### Manual Testing in Staging
The auto-match algorithm was tested against real PagerDuty services in staging, demonstrating accurate matching with confidence scores:

Screenshot 2026-01-13 at 17 33 56

**Test Results:**
- Successfully matched services with various naming patterns
- Team matching bonus (+10%) working as expected
- Acronym matching bonus (+5%) working as expected
- Confidence levels accurately reflect match quality

---

## Breaking changes

N/A - This is a new feature with no breaking changes to existing functionality.

---

## Acknowledgement

- [x] I acknowledge that any contribution to this repository is provided under the terms of the Apache 2.0 license.


plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -765,6 +771,117 @@ export async function createRouter(
...
+      // Validate catalog API is available
+      if (!catalogApi) {
+        response.status(500).json({
+          error: 'CatalogApi not available',
+        });
+        return;
+      }
+
+      // Get threshold from request body (default to 80)
+      const threshold: number = request.body.threshold ?? 80;
```
jhfgloria: Why 80% as default?

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -765,6 +771,117 @@ export async function createRouter(
...
+        threshold > 100
+      ) {
+        response.status(400).json({
+          error: 'Invalid threshold. Must be a number between 0 and 100.',
+        });
+        return;
+      }
+
+      // Get optional flag to return only best match per service
+      const bestOnly: boolean = request.body.bestOnly ?? false;
```
jhfgloria: Is this a thing we want?

plugins/backstage-plugin-backend/src/services/dataLoader.ts
```diff
@@ -0,0 +1,203 @@
+/**
```
jhfgloria: I don't think we need all these comments. Altho this is a public distributed package, Backstage plugins work more as applications than libraries. No one will ever read this to be honest, or use in their code, so it is causing more pollution than helping tbh.

plugins/backstage-plugin-backend/package.json
```diff
@@ -30,14 +30,15 @@
...
-    "prepack": "backstage-cli package prepack",
+    "prepack": "node ../../scripts/replace-versions.js backstage-plugin-backend && backstage-cli package prepack",
     "postpack": "backstage-cli package postpack"
   },
   "dependencies": {
     "@backstage/backend-common": "^0.25.0",
     "@backstage/backend-plugin-api": "backstage:^",
     "@backstage/catalog-client": "backstage:^",
     "@pagerduty/backstage-plugin-common": "workspace:~",
+    "@skyra/jaro-winkler": "^1.1.1",
```
jhfgloria, line 41: Can we confirm with FE/AST if it is okay to use this?

jhfgloria (changes requested):
I did not complete the review of the PR. The amount of comments completely disrupts the readability of the code. I understand that it may have been generated with some AI model, but it gets really hard to review the code... :/ Could you remove some of the comments? I don't think we need any of the function annotations because this is not a library, and we don't need those comments before every line of code.

plugins/backstage-plugin-backend/package.json
fanfernandes: https://pagerduty.slack.com/archives/C5SF6HHGV/p1768557136517949

plugins/backstage-plugin-backend/src/service/router.ts
fanfernandes: Good question!
The endpoint returns data for frontend consumption - no auto-creation yet. Since the design isn't fully defined (suggestions vs auto-sync), 80% default provides flexibility and better coverage (64.4% vs 12.8% at 100% in the tests).
I'm open to changing the default to 100% as Sándor originally suggested in the design doc. Users can always lower it via request body if needed.
What do you prefer?

plugins/backstage-plugin-backend/src/service/router.ts
fanfernandes: I think so for flexibility - it reduces noise.
**Without it:** A single PD service might match 5+ Backstage components (all above threshold)
**With it:** Each PD service gets only its highest-scoring match
Useful for getting a clean overview. Optional, defaults to false.
**Question:** Do you know if the frontend will show all possible matches per service, or just the best one? If it's always "show all", I can remove this since frontend can filter if needed.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -765,6 +771,109 @@ export async function createRouter(
...
 
+  // POST /mapping/entity/auto-match
+  router.post('/mapping/entity/auto-match', async (request, response) => {
+    try {
+      if (!catalogApi) {
+        response.status(500).json({
+          error: 'CatalogApi not available',
+        });
+        return;
+      }
```
jhfgloria: I'm not really sure we need to assert this, because `catalogApi` is injected as a dependency in the router. I don't see how could it not be "present". If you really want to do this I would never return a strict `500` because it was caught in runtime and 500 is reserved for when something unexpected happens. I would lean into a 503, altho I don't think I would ever assert here.

plugins/backstage-plugin-backend/src/service/router.ts
jhfgloria: I think the default should in fact be 100%. If customers are confident (or not) they can adjust that threshold, but that's up to them.

plugins/backstage-plugin-backend/src/services/dataLoader.ts
```diff
@@ -0,0 +1,104 @@
...
+  pdServices: NormalizedService[];
+  bsComponents: NormalizedService[];
+}
+
+export async function loadPagerDutyServices(): Promise<NormalizedService[]> {
+  try {
+    const services: PagerDutyService[] = await getAllServices();
+
+    const normalizedServices: NormalizedService[] = services.map(service => {
+      const teamName = service.teams?.[0]?.summary ?? '';
```
jhfgloria, line 32: I didn't know this was a thing. How can a service belong to multiple teams? 🤔

plugins/backstage-plugin-backend/src/services/dataLoader.ts
```diff
@@ -0,0 +1,104 @@
...
+      }`,
+    );
+  }
+}
+
+export async function loadBackstageComponents(
+  context: DataLoaderContext,
+): Promise<NormalizedService[]> {
+  try {
+    const { catalogApi } = context;
```
jhfgloria: You can do this step in the function signature to simplify the code.

```javascript
export async function loadBackstageComponents({
  catalogApi,
}: DataLoaderContext): Promise<NormalizedService[]> {
...
```

plugins/backstage-plugin-backend/src/services/dataLoader.ts
```diff
@@ -0,0 +1,104 @@
...
+        kind: 'Component',
+      },
+    });
+
+    const normalizedComponents: NormalizedService[] = response.items.map(
+      (entity: Entity) => {
+        const entityRef =
+          `${entity.kind}:${entity.metadata.namespace}/${entity.metadata.name}`.toLowerCase();
+
+        const owner = (entity.spec?.owner as string | undefined) ?? '';
```
jhfgloria: What's the point of `as string | undefined` here? That comes explicitly from the `?.` annotation I believe.

plugins/backstage-plugin-backend/src/utils/normalization.ts
```diff
@@ -0,0 +1,106 @@
...
+  sourceId: string;
+  source: 'pagerduty' | 'backstage';
+}
+
+export function normalizeService(
+  rawName: string,
+  teamName: string,
+  sourceId: string,
+  source: 'pagerduty' | 'backstage',
+  useAdvancedPreprocessing: boolean = false,
```
jhfgloria: Whats is this? I don't think we use anywhere so my assumption is _we don't need it_.

plugins/backstage-plugin-backend/src/utils/normalization.ts
```diff
@@ -0,0 +1,106 @@
...
+export interface NormalizedService {
+  rawName: string;
+  normalizedName: string;
+  teamName: string;
+  acronym: string;
+  sourceId: string;
+  source: 'pagerduty' | 'backstage';
+}
+
+export function normalizeService(
```
jhfgloria: I don't really like this type of methods. This is usually called a factory, but it kind of falls apart for me because the naming convention trembles when you use 2 different structures as input. For example when you pass a Backstage service you're sending the internal reference as `sourceId`, while when you pass a PagerDuty service you're sending the `service.id`. I think we should have 2 methods one to normalize a Backstage Service and another to normalize PagerDuty services. You can even drop the `source`.

plugins/backstage-plugin-backend/src/utils/normalization.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  let normalized = name.toLowerCase();
+  normalized = normalized.replace(/_/g, ' ').replace(/-/g, ' ');
+  normalized = normalized.replace(/\s+/g, ' ');
+
+  return normalized.trim();
+}
+
+export function preprocessForMatching(name: string): string {
```
jhfgloria: I don't think we need this right now. It's not even being used from what I can tell.

plugins/backstage-plugin-backend/src/utils/normalization.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  const capitals = name.match(/[A-Z]/g);
+  if (capitals && capitals.length > 1) {
+    return capitals.join('');
+  }
+
+  if (name.length > 0) {
+    return name.charAt(0).toUpperCase();
+  }
```
jhfgloria, line 55: This part of the algorithm is quite fuzzy. I'm not sure I understand the concept or even if it would ever reach this point.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -765,6 +771,109 @@ export async function createRouter(
...
+          highConfidenceMatches: highConfidence,
+          mediumConfidenceMatches: mediumConfidence,
+          threshold,
+          loadTimeMs: loadTime,
+          matchTimeMs: matchTime,
+          totalTimeMs: loadTime + matchTime,
+        },
+      });
+    } catch (error) {
+      logger.error(`❌ Auto-match failed: ${error}`);
```
jhfgloria: Please, remove the emoji from the log message. It can be disruptive for the customer.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -765,6 +771,109 @@ export async function createRouter(
...
+      if (error instanceof HttpError) {
+        response.status(error.status).json({
+          errors: [`${error.message}`],
+        });
+      } else {
+        response.status(500).json({
```
jhfgloria, line 867: 500s are not the caught error, they're more suited for unexplained errors. Do we know how did we end on the `else` clause? It would help defining a better status code.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -765,6 +771,109 @@ export async function createRouter(
...
+          totalPossibleComparisons: totalComparisons,
+          matchesFound: matches.length,
+          exactMatches,
+          highConfidenceMatches: highConfidence,
+          mediumConfidenceMatches: mediumConfidence,
+          threshold,
+          loadTimeMs: loadTime,
+          matchTimeMs: matchTime,
+          totalTimeMs: loadTime + matchTime,
+        },
```
jhfgloria, line 853: Is this necessary? Are we presenting this to the users? 🤔

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  return matches;
+}
+
+export function calculateMatchScore(
+  pdService: NormalizedService,
+  bsComponent: NormalizedService,
+): MatchResult {
+  const exactNameMatch =
+    pdService.normalizedName === bsComponent.normalizedName;
```
jhfgloria, line 55: Shouldn't we also perform the empty name validation like we do with the team name?

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  const teamMatch =
+    pdService.teamName !== '' &&
+    bsComponent.teamName !== '' &&
+    pdService.teamName === bsComponent.teamName;
+
+  const acronymMatch =
+    pdService.acronym !== '' &&
+    bsComponent.acronym !== '' &&
+    pdService.acronym === bsComponent.acronym;
```
jhfgloria: Why are we not applying `jaro winkler` to these ones?

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
```diff
@@ -0,0 +1,179 @@
...
+      baseScore: Math.round(baseScore * 100) / 100,
+      exactMatch: false,
+      teamMatch,
+      acronymMatch,
+      rawScore: Math.round(rawScore * 100) / 100,
+    },
+  };
+}
+
+export function findBestMatch(
```
jhfgloria: Is this ever used?

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
```diff
@@ -0,0 +1,179 @@
...
+      (!bestMatch || matchResult.score > bestMatch.score)
+    ) {
+      bestMatch = matchResult;
+    }
+  }
+
+  return bestMatch;
+}
+
+export function groupMatchesByService(
```
jhfgloria: If this method is used internally in this module, please don't export it.

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  for (const serviceMatches of grouped.values()) {
+    if (serviceMatches.length > 0) {
+      bestMatches.push(serviceMatches[0]);
+    }
+  }
+
+  bestMatches.sort((a, b) => {
+    if (b.score !== a.score) {
+      return b.score - a.score;
```
jhfgloria, line 145: What if you have multiple matches with the high score? 🤔

jhfgloria (commented):
Left you some comments! Looks good at first glance.

plugins/backstage-plugin-backend/src/utils/normalization.ts
fanfernandes: It was an improvement for the future, but you are right not used, lets clean it

jhfgloria: I have a philosophy for things like this: [_YAGNI_](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
fanfernandes: No is not, removed

plugins/backstage-plugin-backend/src/service/router.ts
fanfernandes: Yes it makes sense, removed

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
fanfernandes: Good catch!

plugins/backstage-plugin-backend/src/services/dataLoader.ts
fanfernandes: https://pagerduty.slack.com/archives/C05Q4676R7B/p1768573334031489
For this answer services can have multiple teams (EP team + teams from schedules). But we're using only the first team for simplicity (service.teams?.[0]). I think this works for this work since we only need one team for the matching bonus.

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
fanfernandes: Good question! We followed the original design document specification for this feature:

Design Doc says:
- Jaro-Winkler similarity - returns a number between 0 and 1
- Acronym equality + 0.05
- Team name match + 0.05
The use of "equality" for acronyms and "match" for teams (as opposed to "similarity") indicates these should be exact matches, not fuzzy scoring.

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
fanfernandes: Good question! When a PD service has multiple BS matches with the same score, we rely on stable sort and take the first one. Since this is rare (most services have one clear best match), we kept it simple. We could think on a tie-breaker if needed.

plugins/backstage-plugin-backend/src/utils/normalization.ts
fanfernandes: The extractAcronym function uses a fallback chain with 6 stages to handle different naming patterns:

Stage 1-4 (Primary extraction - tested):
1. Empty check
2. Short all-caps ("API" → "API")
3. Word separators ("auth-service" → "AS")
4. CamelCase boundaries ("AuthService" → "AS")

Stage 5 (Lines 48-51 - the "fuzzy" part you asked about executes on rare edge cases)

Example: "POSTGRESQL" (7 chars, all caps)
- Stage 2 fails (>5 chars)
- Stage 3 fails (no separators)
- Stage 4 fails (no camel case boundaries)
- Stage 5 matches: Extracts all capitals → "POSTGRESQL"

Stage 6 (Lines 53-55 - single character fallback):
When nothing else matches, return first letter uppercased.
Example: "service" → "S"

I've added test cases demonstrating this behaviour.

Does that work for you?

plugins/backstage-plugin-backend/src/service/router.ts
fanfernandes: Looking at the original design doc, the API response specification I think only included:
{
  "totalMatches": 137,
  "threshold": 0.9
}

But the current implementation includes more detailed statistics:
statistics: {
...
}

These extra fields could be useful for:
- UI summary displays ("Found 45 matches out of 686 services")
- Performance monitoring
- Debugging

Also @sandornagy517 said that something can change?!?!

But if the frontend doesn't need them, we can simplify to match the design doc.

Should we:
1. Keep the detailed statistics (useful for potential UI features)
2. Simplify to just matchesFound and threshold (matches design doc)

What's your preference? I'm ok with both

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
jhfgloria: That's fine. That's a valid option. I believe there will be some mismatches in terms of team names, but that's not a big deal I think. And we can tune it later.

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
jhfgloria: Hm, you're assuming there is a clear best match. But what if it has 3 80% matches as best score?

plugins/backstage-plugin-backend/src/service/router.ts
jhfgloria: I'm not saying this information doesn't matter. But by including it in the payload you're sending back to the browser from where the user initially requested. If we're not presenting this data to anyone we should not send it. If the data is for us, maybe we should send it our way, and not to the user.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -765,6 +771,107 @@ export async function createRouter(
...
+      });
+    } catch (error) {
+      logger.error(`Auto-match failed: ${error}`);
+      if (error instanceof HttpError) {
+        response.status(error.status).json({
+          errors: [`${error.message}`],
+        });
+      } else if (error instanceof ServiceLoadError) {
```
jhfgloria, line 861: Okay, looking at the code it seems that this will never happen. At least not during the `loadBothSources` method. Because you're catching whatever error that comes from the private methods you're turning them into a plain `Error`: https://github.com/PagerDuty/backstage-plugins/pull/120/changes#diff-c292268ff0868da5b239af20d2766e6c7b467ac5b78ada6efd559a9f23bdfc15R102

plugins/backstage-plugin-backend/src/services/dataLoader.ts
```diff
@@ -0,0 +1,108 @@
...
+    return {
+      pdServices,
+      bsComponents,
+    };
+  } catch (error) {
+    throw new Error(
+      `Failed to load sources: ${
+        error instanceof Error ? error.message : String(error)
+      }`,
+    );
```
jhfgloria: I would simply remove this and let the error from the 2 methods cascade.

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
```diff
@@ -0,0 +1,160 @@
...
+  if (acronymMatch) {
+    rawScore += 5;
+  }
+
+  const finalScore = Math.min(rawScore, 100);
+
+  return {
+    pagerDutyService: pdService,
+    backstageComponent: bsComponent,
+    score: Math.round(finalScore * 100) / 100,
```
jhfgloria: This operation is not obvious to me... Lets put this scenario: your score is `87.53`
What this does is:
- Multiply by 100: `87.53 * 100 = 8753`
- Rounding: `round(8753) = 8753`
- Divide by 100: `8753 / 100 = 87.53`

Which is itself _the same_ floating number. If you're telling me that the number could be `87.5398` then it makes more sense, but if we're presenting integer as score, why not simply `round(finalScore)`?

jhfgloria (commented):
Left a couple more comments, but the rest looks fine.

plugins/backstage-plugin-backend/src/utils/normalization.ts
fanfernandes: Done! Split into two separate functions as you suggested:

      normalizePagerDutyService(name: string, teamName: string, serviceId: string)
      normalizeBackstageComponent(name: string, owner: string, entityRef: string)

plugins/backstage-plugin-backend/src/service/router.ts
fanfernandes: You're right, the current frontend is just a temporary test component that will be removed before merge.

I'm not sure if the production UI will need these statistics or not. If the final implementation won't display them to users, we should remove them from the API response or only log them internally for monitoring.

Should I:
1. Remove the statistics object entirely
2. Should we Log it internally or no need?
3. Keep just matchesFound and threshold (minimal info in the design doc)

What do you think?

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
fanfernandes: We just pick the first one in the array, which is essentially random/non-deterministic, the user doesn't know why match 1 was chosen over 2 and 3, it's not transparent

Potential Solutions:
1. Add better tie-breaking logic (e.g., alphabetical by BS component name)
2. Prefer matches with bonuses (team match or acronym match)
3. Return ALL tied matches and let the user/UI decide
4. Accept it as "good enough"

What do you think? Should we improve the tie-breaking, or is the current behaviour acceptable?

plugins/backstage-plugin-backend/src/service/router.ts
fanfernandes: Done

plugins/backstage-plugin-backend/src/services/dataLoader.ts
fanfernandes: done

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
fanfernandes: Changed to integer score

jhfgloria (approved):
LGTM! I don't think we need that Test Component, so we can get rid of it, but for me apart of it, it is good.

plugins/backstage-plugin-backend/src/services/dataLoader.test.ts
```diff
@@ -0,0 +1,527 @@
...
+      new Error('PagerDuty API error'),
+    );
+
+    await expect(loadPagerDutyServices()).rejects.toThrow(
+      'Failed to load PagerDuty services',
+    );
+  });
+
+  it('handles large number of services', async () => {
+    // Generate 686 mock services (typical staging environment)
```
joaovgsantos, line 178: is 686 really "large"? 

image

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
```diff
@@ -0,0 +1,160 @@
...
+  const jaroWinklerSimilarity = jaroWinkler(
+    pdService.normalizedName,
+    bsComponent.normalizedName,
+  );
+  const baseScore = jaroWinklerSimilarity * 100;
+
+  const teamMatch =
+    pdService.teamName !== '' &&
+    bsComponent.teamName !== '' &&
+    pdService.teamName === bsComponent.teamName;
```
joaovgsantos: don't already have this?

plugins/backstage-plugin-backend/src/services/dataLoader.test.ts
fanfernandes: I changed the wording, agree its not large

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
fanfernandes: Yes, now calculated once at the top and reused

joaovgsantos: I was talking about teamMatch :P (but good catch the acronym one :)

plugins/backstage-plugin-backend/src/services/dataLoader.test.ts
joaovgsantos: but shouldn't we test with larger numbers? don't have to be a unit Test, but we have an O^2 algorithm, we should test how it behaves with large numbers

plugins/backstage-plugin-backend/src/services/matchingEngine.ts
fanfernandes: teamMatch updated

plugins/backstage-plugin-backend/src/services/dataLoader.test.ts
jhfgloria: Yeah I think we NEED to be testing on the order of 4000 services (I heard this number on a customer call).

jhfgloria: cc @fanfernandes

fanfernandes: Ran local performance test:
  - Works fine for PDT-Circular (685 services)
  - For 16,681 services: ~2.4 minutes, manageable but high memory (2.3GB) (tested locally)
I think I should keep it simple for now. What do you think?

jhfgloria: 2.3GB in memory? 😬 I think we may have to:
1. improve that
2. benchmark it properly and provide numbers in the public docs.

There are customers that self host it with Backstage in smaller machines that may not be able to handle it. And I'm not sure Portal instances are ready for it... 🤔

fanfernandes: > Yeah I think we NEED to be testing on the order of 4000 services (I heard this number on a customer call).
Should we do that in a unit test? @jhfgloria

jhfgloria: I'm not worried about the time. This is not a crucial operation on the day to day.

fanfernandes: I did locally a unit test and passed with this results: 
 passed in 29 seconds for 4000 services × 1000 components (4M comparisons).

fanfernandes: should we keep that way?

jhfgloria: What about memory on that one?

fanfernandes: let me do some more realistic tests locally

joaovgsantos (approved):
approving, but we need to be careful on the big number of services associated with a low threshold

Issue #19, opened 2024-11-13
Mapping is not cleaned up in database when service is removed from PagerDuty


aladjadj:
**Describe the bug**
If a service is deleted or recreated on PagerDuty side, the mapping was not cleaned up

**To Reproduce**
1. Create a service in PagerDuty
2. Create a service  in Backstage (Without pagerduty annotation)
3. Map the service in backstage instance /pagerduty
4. The service is Sync and the Backstage service has new pagerduty annotations
-> At this step all good <-
5. Delete the service in the PagerDuty
6. Re-create the service in the PagerDuty (with new ID)
7. The new PagerDuty service is available in /pagerduty but it is impossible to map the Backstage Service again.

**Expected behavior**
- The service from backstage (mapped to old pagerduty service) should be available to re-assign to new service.
- Add the capability to delete existing mapping

**Screenshots**
<img width="1261" alt="image" src="https://github.com/user-attachments/assets/ed62974d-1c3e-400e-b9ef-8cee9dde6597">

PR #128, opened 2026-01-20
fix: Mapping is not cleaned up in database when service is removed from PagerDuty

pdt-rmunoz:
### Description

This PR adds functionality to manually unmap a PagerDuty service from a Backstage entity when the service has been deleted or is no longer accessible in PagerDuty.

**Key changes:**
  - Added "Unmap Service" button in the ServiceNotFoundError component
  - Implemented confirmation dialog with service details before unmapping
  - Added `getEntityMapping()` and `removeServiceMapping()` API methods
  - Enhanced mock data to test deleted service scenarios

image

**Issue number:** #19 

### Affected plugin

- [x] backstage-plugin

### Type of change

### Checklist

- [x] Changes are documented

## Acknowledgement


plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
```diff
@@ -140,7 +163,14 @@ export const PagerDutyCard = (props: PagerDutyCardProps) => {
         errorNode = <MissingTokenError />;
         break;
       case NotFoundError:
-        errorNode = <ServiceNotFoundError />;
+        errorNode = (
+          <ServiceNotFoundError
+            entity={entity}
+            serviceId={props.serviceId}
+            integrationKey={props.integrationKey}
+            onUnmap={entity ? handleUnmapService : undefined}
```
jhfgloria: At this point would the `entity` ever be nullable? 🤔

plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
```diff
@@ -95,6 +108,16 @@ export const PagerDutyCard = (props: PagerDutyCardProps) => {
     setRefreshStatus(x => !x);
   }, []);
 
+  const handleUnmapService = useCallback(async () => {
+    if (!entity) return;
```
jhfgloria: Same, would this ever happen? 🤔

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
-    action={
-      <Button
-        color="primary"
-        variant="contained"
-        href="https://pagerduty.github.io/backstage-plugin-docs/getting-started/pagerduty/"
-      >
-        Read More
-      </Button>
+export interface ServiceNotFoundErrorProps {
+  /** The entity that has the orphaned mapping */
```
jhfgloria: I don't think we need this type of comments. The code is self explanatory. 👍

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
+    if (!onUnmap) return;
+    return await onUnmap();
+  });
+
+  useEffect(() => {
+    if (value !== undefined) {
+      (async () => {
+        alertApi.post({
+          message: 'Service successfully unmapped',
+        });
```
jhfgloria, line 63: Does this work in Portal? 🤔

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
+
+        setShowConfirmDialog(false);
+
+        // Wait before reloading page
+        await new Promise(resolve => setTimeout(resolve, 1000));
```
jhfgloria: Why wait?

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
+  integrationKey,
+}: ServiceNotFoundErrorProps) => {
+  const [showConfirmDialog, setShowConfirmDialog] = useState(false);
+  const alertApi = useApi(alertApiRef);
+  const theme = useTheme();
+
+  const [{ value, loading, error }, handleUnmap] = useAsyncFn(async () => {
+    if (!onUnmap) return;
+    return await onUnmap();
+  });
```
jhfgloria: Shouldn't onUnmap be a dependency of this hook, or it does not support dependencies? 🤔

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
+  serviceId,
+  integrationKey,
+}: ServiceNotFoundErrorProps) => {
+  const [showConfirmDialog, setShowConfirmDialog] = useState(false);
+  const alertApi = useApi(alertApiRef);
+  const theme = useTheme();
+
+  const [{ value, loading, error }, handleUnmap] = useAsyncFn(async () => {
+    if (!onUnmap) return;
+    return await onUnmap();
```
jhfgloria: I think you can rewrite like this in TS.
```suggestion
   return await onUnmap?.();
```

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
+        setShowConfirmDialog(false);
+
+        // Wait before reloading page
+        await new Promise(resolve => setTimeout(resolve, 1000));
+        window.location.reload();
+      })();
     }
-  />
-);
+  }, [value, alertApi]);
```
jhfgloria: `setShowConfirmDialog` should be in the dependencies too.

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
+              variant="contained"
+              href="https://pagerduty.github.io/backstage-plugin-docs/getting-started/pagerduty/"
+            >
+              Read More
+            </MuiButton>
+          </Box>
+        }
+      />
+
+      {/* Confirmation Dialog */}
```
jhfgloria: nit: We don't need this comment. We can refactor this into its own component, otherwise the less comments the better.

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
+      >
+        <DialogTitle>Unmap PagerDuty Service?</DialogTitle>
+        <DialogContent>
+          <Alert severity="info">
+            <Typography variant="body1" align="justify">
+              If this is a temporary issue or the service still exists in
+              PagerDuty, please verify the service ID in your entity
+              configuration before unmapping. If the service has been
+              permanently deleted or renamed, unmapping will allow you to
+              create a new mapping to the correct service.
```
jhfgloria, line 124: Lets just confirm this copy with @pdt-bayon and Derek.

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
```diff
@@ -14,22 +14,188 @@
...
+            data-testid="unmap-button"
+            id="unmap"
+            style={{
+              background: loading
+                ? theme.palette.action.disabled
+                : theme.palette.secondary.main,
+            }}
+            isDisabled={loading}
+            variant="primary"
+            onClick={() => handleUnmap()}
```
jhfgloria: This is a registered callback no? You could simply pass the reference, otherwise you lose the benefits of wrapping it. When you type `() => handleUnmap()` you're recreating a new reference everytime, which could lead to unnecessary rerenders (not sure if it would be the case). You can simply:

```suggestion
            onClick={handleUnmap}
```

plugins/backstage-plugin/src/api/client.ts
```diff
@@ -174,6 +174,45 @@ export class PagerDutyClient implements PagerDutyApi {
...
 
+  async getEntityMapping(entityRef: string): Promise<{
+    mapping: {
+      serviceId: string;
+      integrationKey: string;
+      entityRef: string;
+      account: string;
+    };
+  }> {
+    // Parse entity reference
```
jhfgloria: nit: remove.

plugins/backstage-plugin/src/api/client.ts
```diff
@@ -174,6 +174,45 @@ export class PagerDutyClient implements PagerDutyApi {
...
+  async removeServiceMapping(entityRef: string): Promise<boolean> {
+    // First, get the current mapping to retrieve serviceId and integrationKey
+    const { mapping } = await this.getEntityMapping(entityRef);
+
+    // Then call storeServiceMapping with empty entityRef to unmap
+    // This is the SAME approach used in MappingTable.tsx line 200-204
+    await this.storeServiceMapping(
+      mapping.serviceId,
+      mapping.integrationKey,
+      '', // Empty string = unmap (same as "None" option in admin page)
```
jhfgloria, line 208: What are the effects of this on the Mappings Table? (this will change once @sandornagy517 PR lands, because it inverts the mappings, but until then.)

jhfgloria (commented):
Left some comments and suggestions. Nice Work! 👍

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: API response is too fast, so we need it to see the message it was unmapped

plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
pdt-rmunoz: Good question! While entity is always present on entity pages, it's optional in tests and some components like HomePagePagerDutyCard and PagerDutySmallCard. I've kept it optional to maintain backwards compatibility and avoid breaking 14+ test cases.

plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
pdt-rmunoz: The defensive checks are necessary for type safety

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: removed

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: no, I fixed it!

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: I modified the message but it could be removed

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: removed

plugins/backstage-plugin/src/api/client.ts
pdt-rmunoz: removed

plugins/backstage-plugin/src/api/client.ts
pdt-rmunoz: no effect according what we discuss in our daily meeting

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: fixed

plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
jhfgloria: Well, I don't think there will be a backward compatibility because there are no developers using this code other than us. And you can't be "in-between" versions. Altho Backstage plugins are distributed as "libraries" they are applicational code. If I understand correctly the `entity` is injected by us [here](https://github.com/PagerDuty/backstage-plugins/pull/128/changes/BASE..547fc96b2847407a8587aff27cfdb260924c2960#diff-8873f691138978444dc55f4fdeb5356cd232503c66220f4ad4fdbdfdc58f38f4R40), and it does not change the top-level component signature (e.g. `EntityPagerDutyCard`) being imported and used by customers. Last but not least, we can in fact make breaking changes, as long as we announce them. Updating versions should be a customer's manual process.

plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
jhfgloria: Well, that's only true because you annotate it as `optional` 😁

jhfgloria (commented):
Left you some extra comments. Could you include a changeset for your change? You would need to run `yarn changeset`. I think your changes require it to be a minor instead of a patch because there's a fix through new behaviour.

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: removed!

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: this was fixed, please review

plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
pdt-rmunoz: lol, you are right! not defensive check here

plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
pdt-rmunoz: You were right - since entity is injected internally, I made it required everywhere. Fixed PagerDutyCard, ServiceNotFoundError, HomePagePagerDutyCard, PagerDutySmallCard, and all their tests (22 test cases total). No more optional entity!

jhfgloria (approved):
LGTM 🚀

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: solved

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: fixed

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-rmunoz: true, fixed

plugins/backstage-plugin/src/components/PagerDutyCard/index.tsx
pdt-rmunoz: Solved

plugins/backstage-plugin/src/components/Errors/ServiceNotFoundError.tsx
pdt-bayon: LGTM

PR #134, opened 2026-02-04
Add auto mapping dialog

sandornagy517:
### Description
Add auto mapping dialog.

### Affected plugin

- [X] backstage-plugin

### Type of change

### Checklist

- [ ] I have performed a self-review of this change
- [ ] Changes are documented
- [ ] Changes generate _no new warnings_
- [ ] PR title follows [conventional commit semantics](https://www.conventionalcommits.org/en/v1.0.0/)

## Acknowledgement


examples/entities.yaml
```diff
@@ -53,3 +53,3603 @@ spec:
     message ExampleMessage {
       string example = 1;
     };
+---
```
sandornagy517, line 56: Added these so we have better ways to see the data in development, not sure if we should keep it or revert back to the original 2 services. @jhfgloria what do you think?

packages/app/src/theme/tokens.css
```diff
@@ -0,0 +1,4 @@
+:root {
+  --bui-bg-solid: #00a67e;
+  --bui-bg-solid-hover: #008f6d;
+}
```
jhfgloria, line 4: What's the goal here?

plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
```diff
@@ -78,6 +79,32 @@ export class PagerDutyBackendDatabase implements PagerDutyBackendStore {
...
   }
 
+  async bulkInsertEntityMappings(
+    entities: PagerDutyEntityMapping[],
+  ): Promise<string[]> {
+    if (entities.length === 0) {
+      return [];
+    }
+
+    const now = new Date();
```
jhfgloria: `processedDate` has default value (CURRENT_TIMESTAMP). You shouldn't need to do this if you're bulk inserting.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -651,6 +651,159 @@ export async function createRouter(
...
+        });
+        return;
+      }
+
+      if (mappings.length === 0) {
+        response.status(400).json({
+          error: "Bad Request: 'mappings' array cannot be empty",
+        });
+        return;
+      }
```
jhfgloria: I wouldn't handle this as an error. If its empty there's no insertions.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -651,6 +651,159 @@ export async function createRouter(
...
+        }
+
+        if (existingServiceIds.has(entity.serviceId)) {
+          // Mapping already exists, discard it
+          skipped.push(entity);
+          continue;
+        }
+
+        // in case a mapping is defined and no integration exists,
+        // we need to create one
```
jhfgloria: Remove these inline comments please... Their not necessary.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -651,6 +651,159 @@ export async function createRouter(
     }
   });
 
+  // POST /mapping/entities/bulk
+  router.post('/mapping/entities/bulk', async (request, response) => {
```
jhfgloria, line 655: We don't need to do this now, but I feel these controllers are getting huge, which make code really hard to read. We should consider refactoring stuff using more modules. But don't need to it in this PR.

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/hooks/useAutoMatchResults.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface AutoMatchResult {
+  score: number;
+  serviceId: string;
+  account: string;
+  serviceName: string;
+}
+
+export type AutoMatchResults = Record<string, AutoMatchResult>;
+
+export function useAutoMatchResults() {
```
jhfgloria: This seems so unnecessary. This is `useState` with fancy methods, I think this could be totally manageable using just `useState` directly on the component.

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/hooks/useConfirmMappings.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const confirmMappings = async () => {
+    try {
+      setIsConfirming(true);
+      const totalCount = Object.keys(autoMatchResults).length;
+      const mappingsToCreate = Object.entries(autoMatchResults).map(
+        ([entityName, matchData]) => {
+          const entity = mappingEntities?.find(
+            (e: BackstageEntity) => e.name === entityName,
+          );
```
jhfgloria, line 38: Why is mappingEntities optional? And how would there be match results without the mapping entities? Is this because the entity may be mapped already?

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/hooks/useConfirmMappings.ts
```diff
@@ -0,0 +1,94 @@
...
+        },
+      );
+
+      const response = await pagerDutyApi.storeBulkServiceMappings(
+        mappingsToCreate,
+      );
+
+      if (response.ok) {
+        const result = await response.json();
+        const successCount = result.successCount || result.success?.length || 0;
```
jhfgloria: Again, this seems to me over zealous code. In which case is the `successCount` falsy and the `success.length` not?

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/hooks/useConfirmMappings.ts
```diff
@@ -0,0 +1,94 @@
...
+        const errorCount = result.errorCount || 0;
+
+        const counts: MappingCounts = {
+          created: successCount,
+          skipped: skippedCount,
+          errored: errorCount,
+        };
+
+        onClear();
+        onSuccess(successCount, totalCount, counts);
```
jhfgloria: Maybe the `onSuccess` should clear..?

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/hooks/useMappingToast.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export type ToastSeverity = 'success' | 'error';
+
+export interface MappingCounts {
+  created?: number;
+  skipped?: number;
+  errored?: number;
+}
+
+export function useMappingToast() {
```
jhfgloria: This seems very unnecessary. All those things seem like properties that should be passed to a toast... which should be managed by the parent component. Using hooks to remove clutter from the component is an anti-pattern. You remove clutter from a component by building more components. You use hooks to share common behaviour across components. This hook only works for this specific scenario. If you want something managing the mapping toast why not creating the ManagingToast component?

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/AutoMappingsButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+      Auto-Mapping
+      <div
+        style={{
+          backgroundColor: 'var(--bui-bg-solid)',
+          color: 'white',
+          fontSize: '11px',
+          fontWeight: 'bold',
+          padding: '2px 6px',
+          borderRadius: '4px',
+        }}
```
jhfgloria: Avoid `style` please.

jhfgloria (commented):
Left you some comments and questions.

packages/app/src/theme/tokens.css
sandornagy517: In the designs we have this green color for main buttons, I thought I'd add it as a global rule

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/hooks/useConfirmMappings.ts
sandornagy517: It's optional because before loading it can be undefined. We can have match results without the entity because it might load earlier than the mappings or vice-versa

plugins/backstage-plugin-backend/src/service/router.ts
sandornagy517: I agree!

packages/app/src/theme/tokens.css
jhfgloria: Hm. we should revisit that with @derek probably, since the green comes from the design system in Portal. 🤔

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
```diff
@@ -34,24 +42,36 @@ export default function MappingsTable() {
...
-    showError,
-    closeToast,
-  } = useMappingToast();
+  const [toastOpen, setToastOpen] = useState(false);
+  const [toastSeverity, setToastSeverity] = useState<ToastSeverity>('success');
+  const [toastMessage, setToastMessage] = useState('');
+  const [toastTotalMatches, setToastTotalMatches] = useState<number>(0);
+  const [toastMappingCounts, setToastMappingCounts] = useState<MappingCounts>(
+    {},
+  );
```
jhfgloria, line 53: 12 lines and an entire file are now 5+3+3+7 lines of code... (I hate hooks)

jhfgloria (approved):
I believe you changed most of my comments! LGTM!

PR #135, opened 2026-02-06
feat: Add filtering capabilities to service mapping table

jhfgloria:
### Description

This PR implements filtering capabilities for the service mapping table. Users can now filter the mapping table by:
- **Entity Name**: Full-text search via Backstage Catalog API
- **Team Name**: Full-text search via Backstage Catalog API
- **Account**: Match via entity mappings
- **PagerDuty Service Name**: Query PagerDuty API and match via entity mappings
- **Status**: Filter by InSync, OutOfSync, NotMapped, or Error states

The filter UI is hidden by default and can be toggled using a filter icon button. When filters are applied, pagination automatically resets to the first page.

image

image

### Affected plugin

- [x] backstage-plugin

### Type of change

### Checklist

- [x] Changes are documented

### Notes
- [x] This code was AI assisted
- Sorting functionality deferred to a future iteration

## Acknowledgement

### DIFF
--- a/plugins/backstage-plugin-backend/src/apis/pagerduty.ts
+++ b/plugins/backstage-plugin-backend/src/apis/pagerduty.ts
@@ -907,6 +907,87 @@ export async function getAllServices(): Promise<PagerDutyService[]> {
   return allServices;
 }
 
+export async function getServicesByPartialName(
+  partialName: string,
+): Promise<PagerDutyService[]> {
+  const allServices: PagerDutyService[] = [];
+
+  await Promise.all(
+    Object.entries(EndpointConfig).map(async ([account, _]) => {
+      let response: Response;
+      const params = `query=${encodeURIComponent(partialName)}&time_zone=UTC&include[]=integrations&include[]=escalation_policies&include[]=teams&total=true`;
+
+      const token = await getAuthToken(account);
+
+      const options: RequestInit = {
+        method: 'GET',
+        headers: {
+          Authorization: token,
+          Accept: 'application/vnd.pagerduty+json;version=2',
+          'Content-Type': 'application/json',
+        },
+      };
+
+      const apiBaseUrl = getApiBaseUrl(account);
+      const baseUrl = `${apiBaseUrl}/services`;
+
+      let offset = 0;
+      const limit = 50;
+      let result: PagerDutyServicesAPIResponse;
+
+      try {
+        do {
+          const paginatedUrl = `${baseUrl}?${params}&offset=${offset}&limit=${limit}`;
+
+          response = await fetchWithRetries(paginatedUrl, options);
+
+          if (response.status >= 500) {
+            throw new HttpError(
+              `Failed to get services. PagerDuty API returned a server error. Retrying with the same arguments will not work.`,
+              response.status,
+            );
+          }
+
+          switch (response.status) {
+            case 400:
+              throw new HttpError(
+                'Failed to get services. Caller provided invalid arguments.',
+                400,
+              );
+            case 401:
+              throw new HttpError(
+                'Failed to get services. Caller did not supply credentials or did not provide the correct credentials.',
+                401,
+              );
+            case 403:
+              throw new HttpError(
+                'Failed to get services. Caller is not authorized to view the requested resource.',
+                403,
+              );
+            default: // 200
+              break;
+          }
+
+          result = (await response.json()) as PagerDutyServicesAPIResponse;
+
+          // set account
+          result.services.forEach(service => {
+            service.account = account;
+          });
+
+          allServices.push(...result.services);
+
+          offset += limit;
+        } while (offset < result.total!);
+      } catch (error) {
+        throw error;
+      }
+    }),
+  );
+
+  return allServices;
+}
+
 export async function getChangeEvents(
   serviceId: string,
   account?: string,
--- a/plugins/backstage-plugin-backend/src/apis/pagerduty.ts
+++ b/plugins/backstage-plugin-backend/src/apis/pagerduty.ts
@@ -889,7 +889,6 @@ export async function getAllServices(): Promise<PagerDutyService[]> {
 
           result = (await response.json()) as PagerDutyServicesAPIResponse;
 
-          // set account
           result.services.forEach(service => {
             service.account = account;
           });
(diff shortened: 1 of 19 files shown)


plugins/backstage-plugin-backend/src/services/pagerduty/index.ts
```diff
@@ -0,0 +1,12 @@
+import { getAllServices } from "../../apis/pagerduty";
+
+export async function getServicesIdsByPartialName(partialName: string): Promise<string[]> {
+  const pagerdutyServices = await getAllServices();
```
fanfernandes: `getServicesIdsByPartialName` calls `getAllServices() `on every filtered request. Is there a way we can avoid that?

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
```diff
@@ -0,0 +1,443 @@
...
+          kind: string;
+          'metadata.namespace'?: string;
+          'metadata.name'?: string;
+        }>;
+        limit: number;
+        offset: number;
+        fullTextFilter?: { term: string; fields: string[] };
+      } = {
+        filter: [{ kind: 'Component' }],
+        limit: needsPostProcessing ? 10000 : limit,
```
fanfernandes: Why 10000?

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
```diff
@@ -0,0 +1,443 @@
...
+            return;
+          }
+        } else {
+          response.json({ entities: [], totalCount: 0 });
+          return;
+        }
+      }
+
+      const componentEntities = await _queryEntitiesWithFilters(
+        catalogApi!,
```
fanfernandes: `catalogApi!` — should we guard for `undefined` or change the type to `CatalogApi` instead of `CatalogApi | undefined`?

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
```diff
@@ -0,0 +1,443 @@
...
+import * as Pagerduty from '../services/pagerduty';
+import * as CatalogEntityUtils from '../utils/catalog-entity';
+import { PagerDutyBackendStore } from '../db';
+import { CatalogApi, GetEntitiesResponse, QueryEntitiesResponse } from '@backstage/catalog-client';
+import { HttpError, PagerDutyEntityMappingsResponse, PagerDutyService } from '@pagerduty/backstage-plugin-common';
+import { getServiceByIntegrationKey, getServicesByIds } from '../apis/pagerduty';
+import { RawDbEntityResultRow } from '../db/PagerDutyBackendDatabase';
+
+export function getMappingEntities(store: PagerDutyBackendStore, catalogApi: CatalogApi | undefined) {
+  return async function getMappingEntitiesFunction(request: Request, response: Response) {
```
sandornagy517, line 60: why the named function here?

sandornagy517 (commented):
the frontend code looks great! I'll still review the backend part later!!

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
jhfgloria: I can use an arrow-function if you'd prefer. Being a regular function the linter wanted to be named (and I also like to do this, because stack traces will use it instead of `anonymous()`).

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
jhfgloria: Yeah I'll change the signature to stop supporting the `undefined` value. 👍

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
jhfgloria: We need to setup something I guess. When post processing is needed we don't how many items will be needed. I can change to _not have any limit_ if it needs post processing if you find a better approach.

plugins/backstage-plugin-backend/src/services/pagerduty/index.ts
jhfgloria: That's a nice catch, for sure, maybe we can try to cache it on static class service, what do you think? Altho the number of services can be huge and the cache would be shared across accounts... 🤔 This way it is eventually garbage collected.

fanfernandes: Good point about the cache complexity. PagerDuty services API supports a query parameter for server-side filtering, it's already used in this same file by getServiceByIntegrationKey. Could we create a similar function that passes the partialName as the query param instead of fetching all services? 
https://developer.pagerduty.com/api-reference/e960cca205c0f-list-services

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
fanfernandes: Makes sense. Could we extract this to a named constant e.g., MAX_ENTITIES_FOR_POST_PROCESSING I think its more clear? but a nit :)

fanfernandes (approved):

plugins/backstage-plugin-backend/src/services/pagerduty/index.ts
jhfgloria: I was the one proposing the use of that method. The issue here is that the filtered request does not allow for partial values... which is the entire goal. Looking for `service` should return `Backstage service`, `Webhook service` and so on.

sandornagy517 (approved):
great job!!!

plugins/backstage-plugin-backend/src/services/pagerduty/index.ts
jhfgloria: Nevermind, you're right. I can use the `query` parameter in that request to do partial name search!

plugins/backstage-plugin-backend/src/apis/pagerduty.ts
```diff
@@ -907,6 +907,87 @@ export async function getAllServices(): Promise<PagerDutyService[]> {
...
+                'Failed to get services. Caller is not authorized to view the requested resource.',
+                403,
+              );
+            default: // 200
+              break;
+          }
+
+          result = (await response.json()) as PagerDutyServicesAPIResponse;
+
+          // set account
```
sandornagy517: the comment feels redundant

sandornagy517 (approved):
great job!

PR #143, opened 2026-03-04
feat: Add match score sorting

sandornagy517:
### Description

Add sorting for match scores in the mappings table.

### Affected plugin

- [X] backstage-plugin

### Type of change

### Checklist

- [ ] Changes are documented

## Acknowledgement


plugins/backstage-plugin/src/components/PagerDutyPage/AutomaticMappingsDialog.tsx
```diff
@@ -28,6 +28,11 @@ interface AutomaticMappingsDialogProps {
         serviceId: string;
         account: string;
         serviceName: string;
+        entity?: {
```
jhfgloria, line 31: Why was this needed on this PR?

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
```diff
@@ -31,71 +31,151 @@ export interface AutoMatchResult {
...
+  };
+  onFilterChange: (
+    key: 'name' | 'serviceName' | 'status' | 'teamName',
+    value: string,
+  ) => void;
+  onEditEntity: (entity: BackstageEntity) => void;
+  onRemoveMatch: (entityName: string) => void;
+}
+
+function MappingsTableContent({
```
jhfgloria: Wondering if this shouldn't go into a different file to reduce the context of this file. Usually I like one component per file, unless they're really a small component (e.g. FilterRow is on its own file).

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
```diff
@@ -31,71 +31,151 @@ export interface AutoMatchResult {
...
+            mappingScore: matchResult.score,
+            autoMatchedServiceId: matchResult.serviceId,
+            autoMatchedServiceName: matchResult.serviceName,
+          } as BackstageEntity;
+        });
 
-  const clearMatches = useCallback(() => {
-    setAutoMatchResults({});
-  }, []);
+        if (filter) {
```
jhfgloria: Food for thought. If we are applying sort and filters again here, shouldn't we have a module that is able to do it instead of repeating the entire thing in both places? No need to do it on this iteration, but something we should think of. The logic is complex enough to be repeated.

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
```diff
@@ -191,11 +227,14 @@ export default function MappingsTable() {
         isSortable: true,
         cell: entity => <StatusCell entity={entity} />,
       },
-      {
+    ];
+    
+    if (hasMatches) {
+      baseColumns.push({
```
jhfgloria: So this column only shows up after doing score analysis?

jhfgloria (approved):

jhfgloria (approved):
Approved with comments.

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
```diff
@@ -136,33 +197,8 @@ export default function MappingsTable() {
     },
   });
```
jhfgloria: Claude made a interesting question while reviewing the PR. Doesn't `useTable` support dependencies array? Otherwise this could be re-rendering every time.

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
```diff
@@ -252,6 +292,122 @@ export default function MappingsTable() {
...
+    {},
+  );
+  const hasMatches = Object.keys(autoMatchResults).length > 0;
+  const tableKey = `${selectedAccount}-${hasMatches ? `matches-${Object.keys(autoMatchResults).length}` : 'no-matches'}`;
+
+  const clearMatches = useCallback(() => {
+    setAutoMatchResults({});
+  }, []);
+
+  const setMatches = (results: AutoMatchResults) => {
```
jhfgloria, line 72: This is the only method that is not memoized. Any reason why? Asking, not saying it should be.

sandornagy517: it's because this one we use in the component, the rest we pass as props

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
sandornagy517: i checked the type definition of the `useTable` and it does not take dependencies, but I might be wrong

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
sandornagy517: yep! I hid it for the default view cause it's just an empty column until we calculate the matches

plugins/backstage-plugin/src/components/PagerDutyPage/MappingsTable/MappingsTable.tsx
sandornagy517: just moved it!

plugins/backstage-plugin/src/components/PagerDutyPage/AutomaticMappingsDialog.tsx
sandornagy517: The automatic match results endpoint returns the entity and the service and now I started using this endpoint to load all the entities with the matches, that's why the type had to be adjusted so when the match object's entity is used it was throwing errors without it

PR #158, opened 2026-05-08
fix: Team filter on auto matchings#157

sandornagy517:
### Description
- Automatic matches were not filtered on selected team filters. This PR is fixing that
- Users with lots of PD services would have timeout issues when calculating matches. Moved to a poll based loading

### Affected plugin

- [x] backstage-plugin

### Type of change

- [X] Fix (non-breaking change which fixes an issue)

### Checklist

- [ ] Changes are documented

## Acknowledgement


plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -3456,4 +3462,135 @@ describe('createRouter', () => {
...
+      expect(response.status).toEqual(200);
+      const matchedEntityNames = response.body.matches.map(
+        (m: { backstageComponent: { name: string } }) => m.backstageComponent.name,
+      );
+      expect(matchedEntityNames).toContain('test-component');
+      expect(matchedEntityNames).toContain('component-1');
+      expect(matchedEntityNames).toContain('component-2');
+      expect(matchedEntityNames).toContain('test-component-filtered');
+      expect(matchedEntityNames).toContain('test-default-search');
+      expect(matchedEntityNames).toContain('component-not-mapped');
```
jhfgloria: How are are these all present in the matches if there are only 2 services returned in the getAllServices mock?

renannoliveira (approved):

plugins/backstage-plugin-backend/src/service/router.test.ts
sandornagy517: it's because of the low threshold, they all pass 80% with a certain %

jhfgloria: But if getAllServices never returns them how is there even a comparison happening?

sandornagy517: I'll adjust the matching algorithm in a follow-up PR!

packages/app/package.json
```diff
@@ -39,7 +39,7 @@
     "@backstage/plugin-techdocs-react": "backstage:^",
     "@backstage/plugin-user-settings": "backstage:^",
     "@backstage/theme": "backstage:^",
-    "@backstage/ui": "backstage:^",
+    "@backstage/ui": "^0.14.3",
```
jhfgloria, line 42: Was this the reason the UI broke? If it was we can't really trust their `backstage:^` token to manage the plugin's dependencies.

plugins/backstage-plugin/src/components/PagerDutyPage/AutomaticMappingsDialog.tsx
fanfernandes, line 256: DEVECO-795 asks for both alphabetical order (fixed) and search. @backstage/ui Select supports this natively via a searchable prop. Could you add searchable and optionally searchPlaceholder="Search teams..." to this Select? Tested locally, the dropdown currently shows a plain list with no search input.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -986,6 +931,56 @@ export async function createRouter(
     }
   });
 
+  // POST /mapping/entity/auto-match/start
+  router.post('/mapping/entity/auto-match/start', async (request, response) => {
```
jhfgloria, line 881: What was the point of having asynchronous matches? 🤔 And when should we use this method or `/mapping/entity/auto-match`?

plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -3456,4 +3479,201 @@ describe('createRouter', () => {
...
+
+    it('POST /start returns 202 with a jobId', async () => {
+      const response = await request(app)
+        .post('/mapping/entity/auto-match/start')
+        .send({ threshold: 100 });
+
+      expect(response.status).toEqual(202);
+      expect(typeof response.body.jobId).toBe('string');
+      expect(['pending', 'running']).toContain(response.body.status);
+    });
```
jhfgloria: For the sake of having a good state in cache, you either clean it, or maybe you should use waitForJob in tests where the assertion is `pending`. Since the cache is shared it can conflict between tests. I'd clean the cache between tests, but that's up to you.

plugins/backstage-plugin-backend/src/services/autoMatchJobs.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    matchesFound: 0,
+    exactMatches: 0,
+    highConfidenceMatches: 0,
+    mediumConfidenceMatches: 0,
+    threshold: 100,
+    loadTimeMs: 0,
+    matchTimeMs: 0,
+    totalTimeMs: 0,
+  },
+};
```
jhfgloria, line 20: If you want to expose a global constant to the tests, please freeze it jus to avoid having tests mess with it.
```suggestion
const emptyResult: AutoMatchEntityMappingsResponse = Object.freeze({
  matches: [],
  statistics: {
    totalPagerDutyServices: 0,
    totalBackstageComponents: 0,
    totalPossibleComparisons: 0,
    matchesFound: 0,
    exactMatches: 0,
    highConfidenceMatches: 0,
    mediumConfidenceMatches: 0,
    threshold: 100,
    loadTimeMs: 0,
    matchTimeMs: 0,
    totalTimeMs: 0,
  },
});
```

plugins/backstage-plugin-backend/src/services/autoMatchJobs.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const createInMemoryCache = (): CacheService => {
```
jhfgloria: Have you seen if Backstage export a mock version of their cache?

plugins/backstage-plugin/src/components/PagerDutyPage/AutomaticMappingsDialog.tsx
sandornagy517: ah great catch! did not notice that it got implemented for select components. search was already working by just typing but without an input field! thank you for the catch!!

plugins/backstage-plugin/src/components/PagerDutyPage/AutomaticMappingsDialog.tsx
```diff
@@ -57,67 +59,104 @@ export default function AutomaticMappingsDialog({
...
     },
+    refetchIntervalInBackground: true,
   });
 
+  useEffect(() => {
+    if (jobStatus?.status === 'completed' && jobStatus.result) {
+      handleAutoMatchResult(jobStatus.result);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [jobStatus?.status]);
```
jhfgloria: What exactly do we get from this? Can I explore previous matchups?

jhfgloria (commented):
Leaving some comments. I don't really understand why do we need the caching system (unless we can explore previous jobs, I don't see any need for it).

plugins/backstage-plugin-backend/src/services/autoMatchJobs.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    return job;
+  }
+
+  async get(jobId: string): Promise<AutoMatchJob | undefined> {
+    return (await this.cache.get(key(jobId))) as AutoMatchJob | undefined;
+  }
+
+  private async write(job: AutoMatchJob): Promise<void> {
+    await this.cache.set(key(job.id), job as unknown as Parameters<CacheService['set']>[1], { ttl: this.ttlMs });
```
jhfgloria: Why do we need this cast? 🤔 `as unknown as Parameters<CacheService['set']`

packages/app/package.json
sandornagy517: yes! hardcoding this has fixed it

plugins/backstage-plugin/src/components/PagerDutyPage/AutomaticMappingsDialog.tsx
sandornagy517: Realized that we don't need the effect, got rid of it!

plugins/backstage-plugin-backend/src/service/router.ts
sandornagy517: good catch, totally forgot to clean up after changing the endpoints. Got rid of it!

plugins/backstage-plugin-backend/src/services/autoMatchJobs.test.ts
sandornagy517: The one I found was just for spying

plugins/backstage-plugin-backend/src/services/autoMatchJobs.ts
sandornagy517: the cache would take `JsonValue` but the job has it's own TS interface and they collide here

fanfernandes (approved):
Thank you @sandornagy517 for the changes. LGTM!

jhfgloria (approved):
Just check that team filter thing that Francisco raised. All good.

PR #162, opened 2026-05-15
feat: Push custom fields to PagerDuty during catalog processing

sandornagy517:
### Description

We can create custom fields now, we need to update the PD service with the custom field values during catalog processing. This PR adds that functionality

### Affected plugin

- [ ] backstage-plugin
- [x] backstage-plugin-entity-processor

### Type of change

- [x] Fix (non-breaking change which fixes an issue)

### Checklist

- [x] Changes have been tested in dark theme
- [x] Changes are documented

## Acknowledgement

### DIFF
--- a/plugins/backstage-plugin-backend/src/apis/pagerduty.test.ts
+++ b/plugins/backstage-plugin-backend/src/apis/pagerduty.test.ts
@@ -19,6 +19,7 @@ import {
   getServiceStandards,
   insertAccountConfig,
   setFallbackAccountConfig,
+  setServiceCustomFieldValues,
   updateCustomField,
 } from './pagerduty';
 
@@ -1672,6 +1673,65 @@ describe('PagerDuty API', () => {
             method: 'PUT',
           }),
         );
+      },
+    );
+  });
+
+  describe('setServiceCustomFieldValues', () => {
+    it.each(testInputs)(
+      'PUTs values to /services/:id/custom_fields/values and returns the response',
+      async () => {
+        const responseBody = {
+          custom_fields: [{ id: 'PD123', value: 'team-a' }],
+        };
+        mocked(fetch).mockReturnValue(mockedResponse(200, responseBody));
+
+        const result = await setServiceCustomFieldValues({
+          serviceId: 'PSERV1',
+          request: {
+            custom_fields: [{ id: 'PD123', value: 'team-a' }],
+          },
+        });
+
+        expect(fetch).toHaveBeenCalledTimes(1);
+        expect(fetch).toHaveBeenCalledWith(
+          expect.stringContaining('/services/PSERV1/custom_fields/values'),
+          expect.objectContaining({
+            method: 'PUT',
+            body: JSON.stringify({
+              custom_fields: [{ id: 'PD123', value: 'team-a' }],
+            }),
+          }),
+        );
+        expect(result).toEqual(responseBody);
+      },
+    );
+
+    it.each(testInputs)(
+      'throws HttpError with status 404 when service not found',
+      async () => {
+        mocked(fetch).mockReturnValue(mockedResponse(404, {}));
+
+        await expect(
+          setServiceCustomFieldValues({
+            serviceId: 'MISSING',
+            request: { custom_fields: [{ id: 'PD123', value: 'x' }] },
+          }),
+        ).rejects.toMatchObject({ status: 404 });
+      },
+    );
+
+    it.each(testInputs)(
+      'throws HttpError with status 400 when request invalid',
+      async () => {
+        mocked(fetch).mockReturnValue(mockedResponse(400, { error: 'bad' }));
+
+        await expect(
+          setServiceCustomFieldValues({
+            serviceId: 'PSERV1',
+            request: { custom_fields: [{ id: 'PD123', value: 'x' }] },
+          }),
+        ).rejects.toMatchObject({ status: 400 });
       },
     );
   });
--- a/plugins/backstage-plugin-backend/src/apis/pagerduty.test.ts
+++ b/plugins/backstage-plugin-backend/src/apis/pagerduty.test.ts
@@ -19,7 +19,6 @@ import {
   getServiceStandards,
   insertAccountConfig,
   setFallbackAccountConfig,
-  setServiceCustomFieldValues,
   updateCustomField,
 } from './pagerduty';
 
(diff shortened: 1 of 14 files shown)


plugins/backstage-plugin-backend/src/apis/pagerduty.test.ts
```diff
@@ -1675,4 +1676,63 @@ describe('PagerDuty API', () => {
       },
     );
   });
+
+  describe('setServiceCustomFieldValues', () => {
```
jhfgloria: I don't see any value in these tests, and honestly I could live without them. We're simply calling a function and asserting that the value returned is the value mocked. All the assertions revolved around the `fetch` which is a dependency and not the subject under test. This would be more than ostensively tested in integration tests.

plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
jhfgloria: Maybe its worth creating a local private function `toBackstageCustomField(arg: RawDbCustomFieldRow)` where we do that parse? What is being returned in the boolean? `0` and `1`?

plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
```diff
@@ -276,13 +284,24 @@ export class PagerDutyBackendDatabase implements PagerDutyBackendStore {
...
+    };
   }
 
-  async getAllCustomFields(subdomain: string): Promise<BackstageCustomField[]> {
-    const rawFields = await this.db<RawDbCustomFieldRow>(
-      'pagerduty_custom_fields',
-    ).where('pagerdutySubdomain', subdomain);
+  async getAllCustomFields(
+    subdomain: string,
+    options?: { enabled?: boolean },
```
jhfgloria: Maybe this makes it less prone to undefined errors?

```suggestion
    options: { enabled?: boolean } = {},
```

subrama6 (approved):

plugins/backstage-plugin-backend/src/service/customFieldsController.ts
```diff
@@ -115,14 +120,59 @@ export class CustomFieldsController {
...
+        response.status(204).end();
+        return;
+      }
+
+      try {
+        await setServiceCustomFieldValues({
+          serviceId,
+          request: { custom_fields: values },
+          account: subdomain,
+        });
```
jhfgloria: `setServiceCustomFieldValues` returns values. Why don't we do anything with it?

plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -3985,6 +3985,68 @@ describe('createRouter', () => {
         expect(Array.isArray(response.body.customFields)).toBe(true);
       },
     );
+
+    it('filters out disabled fields when enabled=true', async () => {
+      const filtered = await request(app).get('/custom-fields?enabled=true');
+      expect(filtered.status).toEqual(200);
+      filtered.body.customFields.forEach((f: { pagerdutyCustomFieldEnabled: boolean }) => {
```
jhfgloria, line 3993: Please assert that the body has data. Otherwise the foreach will never run which makes this test misleading.

plugins/backstage-plugin-backend/src/service/router.test.ts
```diff
@@ -3985,6 +3985,68 @@ describe('createRouter', () => {
...
+
+      expect(response.status).toEqual(400);
+    });
+
+    it('forwards values to PagerDuty and returns 204', async () => {
+      mocked(fetch).mockReturnValue(
+        mockedResponse(200, {
+          custom_fields: [{ id: 'PD123', value: 'team-a' }],
+        }),
+      );
```
jhfgloria: I hate that we keep doing this...

plugins/backstage-plugin-common/src/types.ts
```diff
@@ -488,3 +488,19 @@ export type PagerDutyCustomFieldUpdateRequest = {
...
+
+/** @public */
+export type PagerDutyServiceCustomFieldValuesRequest = {
+  custom_fields: PagerDutyServiceCustomFieldValue[];
+};
+
+/** @public */
+export type PagerDutyServiceCustomFieldValuesResponse = {
+  custom_fields: PagerDutyServiceCustomFieldValue[];
+};
```
jhfgloria, line 508: These look the same, why two of them?

plugins/backstage-plugin-entity-processor/src/apis/client.ts
```diff
@@ -625,6 +628,102 @@ export class PagerDutyClient {
...
+        throw new Error(
+          `Failed to get enabled custom fields. API returned a server error.`,
+        );
+      }
+
+      switch (response.status) {
+        case 400:
+          throw new Error(await response.text());
+        case 404:
+          return [];
```
jhfgloria: Any reason why we're ignoring the 404? This seems semantically incorrect.

plugins/backstage-plugin-entity-processor/src/apis/client.ts
```diff
@@ -625,6 +628,102 @@ export class PagerDutyClient {
...
+  ): Promise<void> {
+    if (values.length === 0) return;
+
+    if (this.baseUrl === '') {
+      this.baseUrl = await this.discovery.getBaseUrl('pagerduty');
+    }
+
+    const params = new URLSearchParams();
+    if (account) params.set('account', account);
+    const query = params.toString() ? `?${params.toString()}` : '';
```
jhfgloria: Is this needed? In the previous method you never check if the `toString` returns values.

plugins/backstage-plugin-entity-processor/src/processor/PagerDutyEntityProcessor.ts
```diff
@@ -199,6 +201,11 @@ export class PagerDutyEntityProcessor implements CatalogProcessor {
           entity.metadata.annotations?.['pagerduty.com/service-id'];
 
         if (serviceId) {
+          const account =
+            entity.metadata.annotations?.['pagerduty.com/account'];
+
+          await this.pushCustomFieldValues(entity, serviceId, account);
```
jhfgloria: Okay here is where I mostly disagree on the implementation of this feature. I believe that this processor is growing to large and becoming bloated with every feature that we want to intercept the catalog sync. I believe you can create an adjacent processor to this one and make it run AFTER the sync (to avoid running for entities that were not yet properly synced). I would need to investigate more about it, but I believe you can even set priority amongst processors to decide which runs first.

jhfgloria (changes requested):
Left some comments, but ultimately I set `Request changes` because of the processor part. I think this live in its own processor (even tho it lives in the same npm package).

plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
sandornagy517: yep, it's `1` and `0`. I'll implement the `toBackstageCustomField`. thank you!

plugins/backstage-plugin-entity-processor/src/processor/PagerDutyCustomFieldsProcessor.ts
```diff
@@ -0,0 +1,70 @@
...
+import { extractValueAtPath } from './extractEntityValue';
+import { PagerDutyServiceCustomFieldValue } from '@pagerduty/backstage-plugin-common';
+
+export interface PagerDutyCustomFieldsProcessorOptions {
+  logger: LoggerService;
+  discovery: DiscoveryService;
+  auth: AuthService;
+}
+
+export class PagerDutyCustomFieldsProcessor implements CatalogProcessor {
```
jhfgloria, line 15: Ah man, so much more readable!

jhfgloria (approved):
LGTM

PR #173, opened 2026-06-12
feat: automated cleanup for custom field sync logs

sandornagy517:
### Description

Adds an automated cleanup mechanism for the `pagerduty_custom_field_sync_logs` table so it stays bounded regardless of catalog size or sync frequency (DEVECO-873).

- **Scheduled task** registered via `coreServices.scheduler` (`scope: 'global'` — DB-backed lock, exactly one replica runs it; every 15 min, 10-min timeout, 1-min initial delay).
- **Hybrid limits**: rows older than **30 days** (TTL) are deleted, and a **global cap of 500k rows** trims the oldest beyond it. All thresholds configurable under `pagerDuty.customFieldsSyncLogs.*` (see `config.d.ts`).
- **Batched deletes**: 10k rows per batch via `whereIn(id, subquery)` (Postgres has no `DELETE … LIMIT`), each batch a short transaction with a 50ms yield, capped at 500 batches/run (~5M deletes — ~4× headroom over the worst-case inflow of 15 fields × 5k entities × 1-min syncs). Concurrent sync inserts are never blocked: deletes target the lowest ids while inserts append the highest.
- **New migration**: bare `timestamp` index — TTL deletes would otherwise full-scan (existing indexes lead with `subdomain`/`serviceId`).
- Note: the TTL cutoff is bound as a UTC string because knex's better-sqlite3 driver converts `Date` bindings to epoch numbers, which never match string-stored `CURRENT_TIMESTAMP` values.
- **Frontend**: adds a Refresh button next to Export CSV on the Sync Logs tab (wired to `useTable`'s `reload()`).
- Fixes two pre-existing lint errors (duplicate `describe` title in `pagerduty.test.ts`, explicit `any` in `getSyncLogs`).

**Issue number:** DEVECO-873

### Affected plugin

- [x] backstage-plugin

### Type of change

### Checklist

- [x] Changes are documented

## Acknowledgement

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
+++ b/plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
@@ -89,7 +89,25 @@ export interface PagerDutyBackendStore {
     entityPaths: string[];
     serviceNames: string[];
   }>;
+  cleanupSyncLogs(
+    options: SyncLogCleanupOptions,
+  ): Promise<SyncLogCleanupResult>;
 }
+
+/** @public */
+export type SyncLogCleanupOptions = {
+  olderThanDays: number;
+  maxRows: number;
+  batchSize: number;
+  maxBatchesPerRun: number;
+};
+
+/** @public */
+export type SyncLogCleanupResult = {
+  deletedByAge: number;
+  deletedByCap: number;
+  batchesUsed: number;
+};
 
 /** @public */
 export type SyncLogQueryOptions = CustomFieldSyncLogFilters & {
@@ -471,7 +489,9 @@ export class PagerDutyBackendDatabase implements PagerDutyBackendStore {
           .orderBy('serviceName', 'asc'),
       ]);
 
-    const total = countResult ? Number((countResult as any).count) : 0;
+    const total = countResult
+      ? Number(countResult)
+      : 0;
 
     return {
       logs: logs || [],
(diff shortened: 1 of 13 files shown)


renannoliveira (commented):
A few suggestions from my review — mostly polish and a real bug in the `countResult` count access. All issues are optional or minor except the `Number(countResult)` one which silently produces `NaN` for the `total` field.

plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
```diff
@@ -471,7 +489,9 @@ export class PagerDutyBackendDatabase implements PagerDutyBackendStore {
           .orderBy('serviceName', 'asc'),
       ]);
 
-    const total = countResult ? Number((countResult as any).count) : 0;
+    const total = countResult
+      ? Number(countResult)
```
renannoliveira: **Bug:** `Number(countResult)` coerces the whole object and silently returns `NaN` — the `.count` property access was lost when removing `as any`.

Knex returns `{ count: string | number }` from `.count('* as count').first()`, so this should be:
```ts
const countResult = await baseQuery().count<{ count: string | number }>('* as count').first();
const total = countResult ? Number(countResult.count) : 0;
```
Typing the query via the generic replaces the old `as any` with an explicit contract.

plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
```diff
@@ -481,4 +501,84 @@ export class PagerDutyBackendDatabase implements PagerDutyBackendStore {
...
+    };
+
+    while (result.batchesUsed < maxBatchesPerRun) {
+      const deleted = await deleteBatch(q => q.where('timestamp', '<', cutoff));
+      result.deletedByAge += deleted;
+      result.batchesUsed += 1;
+      if (deleted < batchSize) {
+        break;
+      }
+      await sleep(50);
```
renannoliveira: **Nit:** `50` is a magic number repeated in two places (age loop and cap loop). Extract to a constant so future tuning is obvious:
```ts
const BATCH_SLEEP_MS = 50;
// ...
await sleep(BATCH_SLEEP_MS);
```

plugins/backstage-plugin-backend/src/db/PagerDutyBackendDatabase.ts
```diff
@@ -481,4 +501,84 @@ export class PagerDutyBackendDatabase implements PagerDutyBackendStore {
...
+      return result;
+    }
+
+    const sleep = (ms: number) => new Promise(r => setTimeout(r, ms));
+
+    // Rows are stamped by the DB clock (CURRENT_TIMESTAMP, UTC on SQLite), so
+    // the cutoff is bound as a 'YYYY-MM-DD HH:MM:SS' UTC string: better-sqlite3
+    // stores timestamps as strings of that format and would never match a
+    // numeric Date binding, while Postgres casts the string to a timestamp.
+    // Clock skew between app and DB is irrelevant at multi-day granularity.
```
renannoliveira, line 529: The comment explains the format well. Worth adding one sentence that the cutoff is always computed in UTC so operators querying the table directly from a local-timezone SQL client are not confused by the apparent timestamp values.

plugins/backstage-plugin-backend/src/services/syncLogsCleanup.ts
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    const durationMs = Date.now() - startedAt;
+    logger.info(
+      `Sync log cleanup: deleted ${result.deletedByAge} expired and ${result.deletedByCap} over-cap row(s) in ${result.batchesUsed} batch(es) (${durationMs}ms)`,
+    );
+
+    if (result.batchesUsed >= cleanupConfig.maxBatchesPerRun) {
+      logger.warn(
+        `Sync log cleanup exhausted its batch budget (${cleanupConfig.maxBatchesPerRun} batches of ${cleanupConfig.batchSize}); a backlog remains and will be processed on the next run`,
```
renannoliveira: The message describes what happened but not what to tune. Suggest:
```ts
logger.warn(
  `Sync log cleanup exhausted its batch budget (${cleanupConfig.maxBatchesPerRun} batches of ${cleanupConfig.batchSize}); ` +
  `a backlog remains. To catch up faster, increase pagerDuty.customFieldsSyncLogs.cleanup.maxBatchesPerRun ` +
  `or decrease pagerDuty.customFieldsSyncLogs.cleanup.frequencyMinutes.`,
);
```

plugins/backstage-plugin-backend/src/services/syncLogsCleanup.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        logger,
+        cleanupConfig,
+      }),
+    ).resolves.toBeUndefined();
+
+    expect(logger.error).toHaveBeenCalledWith(
+      expect.stringContaining('database is locked'),
+    );
+  });
+});
```
renannoliveira, line 177: Missing coverage for the path where the age phase deletes everything and the cap phase is a no-op (`deletedByCap: 0`). Suggest adding after the existing tests:
```ts
it('logs correctly when age phase empties the table (no cap deletions)', async () => {
  const cleanupSyncLogs = jest.fn().mockResolvedValue({
    deletedByAge: 50,
    deletedByCap: 0,
    batchesUsed: 1,
  });
  const logger = mockServices.logger.mock();

  await runSyncLogsCleanup({ store: createStore(cleanupSyncLogs), logger, cleanupConfig });

  expect(logger.info).toHaveBeenCalledWith(
    expect.stringContaining('deleted 50 expired and 0 over-cap row(s)'),
  );
  expect(logger.warn).not.toHaveBeenCalled();
});
```

plugins/backstage-plugin-backend/src/plugin.ts
```diff
@@ -52,12 +57,30 @@ export const pagerDutyPlugin = createBackendPlugin({
...
       }) {
         const pagerDutyBackendStore: PagerDutyBackendStore =
           await PagerDutyBackendDatabase.create(await database.getClient(), {
             skipMigrations: false,
           });
 
+        const cleanupConfig = readSyncLogsCleanupConfig(config);
+        if (cleanupConfig.enabled) {
+          await scheduler.scheduleTask({
+            id: 'pagerduty-custom-field-sync-logs-cleanup',
```
renannoliveira: This string literal is duplicated verbatim in `plugin.test.ts` line 14. A typo in one won't be caught by the other. Consider exporting a constant from `syncLogsCleanup.ts`:
```ts
// syncLogsCleanup.ts
export const SYNC_LOGS_CLEANUP_TASK_ID = 'pagerduty-custom-field-sync-logs-cleanup';
```
Then use `SYNC_LOGS_CLEANUP_TASK_ID` in both files.

plugins/backstage-plugin-backend/src/plugin.ts
```diff
@@ -52,12 +57,30 @@ export const pagerDutyPlugin = createBackendPlugin({
...
+            frequency: { minutes: cleanupConfig.frequencyMinutes },
+            timeout: { minutes: 10 },
```
renannoliveira: These two values are invisible to operators — all other scheduler parameters come from `cleanupConfig`. At minimum, named constants make future tuning obvious:
```ts
// syncLogsCleanup.ts
export const SYNC_LOGS_CLEANUP_TIMEOUT_MINUTES = 10;
export const SYNC_LOGS_CLEANUP_INITIAL_DELAY_MINUTES = 1;
```
```ts
// plugin.ts
timeout: { minutes: SYNC_LOGS_CLEANUP_TIMEOUT_MINUTES },
initialDelay: { minutes: SYNC_LOGS_CLEANUP_INITIAL_DELAY_MINUTES },
```

renannoliveira (approved):

PR #174, opened 2026-06-16
refactor: adjust for Backstage 1.51.2 and port PagerDutyPage to @backstage/ui

sandornagy517:
### Description

Upgrades Backstage `1.47.3` → `1.51.2` and adapts the code to the breaking changes that came with it, then ports the PagerDuty advanced-config page off `@backstage/core-components` onto `@backstage/ui`.

**Backstage 1.51.2 upgrade**
- Bump `backstage.json`, the yarn plugin spec/checksum, and `yarn.lock`
- Move `catalogProcessingExtensionPoint` import out of `@backstage/plugin-catalog-node/alpha` (now exported from the package root)
- Drop the removed `variant="gridItem"` prop from `Entity*` cards in `EntityPage.tsx`
- Switch `@backstage/ui` to the managed `backstage:^` version
- Add explicit `defineParams` type annotations in `dev/index.tsx` for the changed blueprint callback signature
- Consolidate the standalone `NavItemBlueprint` into the `PageBlueprint` `title`/`icon` params

**PagerDutyPage port to @backstage/ui**
- Replace `TabbedLayout` with `Tabs`/`TabList`/`Tab`/`TabPanel`, deriving the active tab from the URL via `useRouteRef` so routing to the existing sub-paths (`/service-mapping`, `/custom-fields`, `/settings`) still works; redirect the bare base path to the first tab
- Swap `Page`/`Header`/`Content` for `@backstage/ui` `Header`, `FullPage`, and `Container`
- Replace the service-mapping description `Card` with an `@backstage/ui` `Alert` (same copy)

**PagerDuty service caching and batch insert fixes**
- Add a 20-minute `CacheService` layer into `getAllServices` and `getServiceById` so repeated calls within a request (auto-match, confirm, bulk mapping) hit the cache rather than the PagerDuty API
- Wire the cache through `createAutoMatchRunner` → `loadBothSources` → `loadPagerDutyServices` so auto-match runs also benefit from the shared cache
- Batch `getSerivcesByIdsAndAccount` into 50-id chunks to stay within API gateway URL-length limits
- Chunk `bulkInsertEntityMappings` into 200-row transactions to avoid SQLite's 500-term compound SELECT limit
- Fix test isolation: clear the in-memory cache store in `beforeEach` so cached responses from one test don't leak into the next

image

### Affected plugin

- [x] backstage-plugin
- [x] backstage-plugin-entity-processor

> Also touches `packages/app` (EntityPage card props).

### Type of change

- [x] Fix (non-breaking change which fixes an issue)

### Checklist

- [ ] Changes are documented
- [ ] Changes generate _no new warnings_

## Acknowledgement

🤖 Generated with [Claude Code](https://claude.com/claude-code)


plugins/backstage-plugin/src/alpha/pages.tsx
```diff
@@ -1,13 +1,60 @@
 import { convertLegacyRouteRef } from '@backstage/core-compat-api';
 import { rootRouteRef } from '../routes';
-import { PageBlueprint } from '@backstage/frontend-plugin-api';
+import { PageBlueprint, SubPageBlueprint } from '@backstage/frontend-plugin-api';
```
renannoliveira, line 3: **[CRITICAL] `SubPageBlueprint` is not exported from the installed `@backstage/frontend-plugin-api` v0.13.2 — the entire alpha PagerDuty page crashes at runtime.**

`SubPageBlueprint` resolves to `undefined` at import time. Every `SubPageBlueprint.make(...)` call (lines 21, 35, 49) throws a `TypeError`. Additionally, `PageBlueprint.make` with no `loader` passes `undefined` to `ExtensionBoundary.lazy`, rendering nothing even if the sub-page calls were guarded.

The developer comment on lines 13–16 describes intended Backstage behavior that is not present in v0.13.2. This needs either a bump to a Backstage release that exports `SubPageBlueprint`, or a revert to a `loader`-based approach.

plugins/backstage-plugin/src/alpha/plugin.ts
```diff
@@ -31,7 +35,9 @@ export const pagerDutyPlugin = createFrontendPlugin({
     pagerDutyEntitySmallCard,
     pagerDutyApi,
     pagerDutyPage,
```
renannoliveira, line 37: **[HIGH] No `NavItemBlueprint` is registered — the PagerDuty sidebar entry is missing in the alpha frontend.**

`nav-items.tsx` was deleted and its `NavItemBlueprint` was not replaced. `PageBlueprint`'s params schema (`path`, `loader`, `routeRef`) does not include `title` or `icon` — those props in `pages.tsx` are silently ignored. `NavItemBlueprint` is a distinct blueprint with kind `"nav-item"` that must be explicitly created and added here.

Alpha frontend users will find no PagerDuty entry in the sidebar.

plugins/backstage-plugin-backend/src/apis/pagerduty.ts
```diff
@@ -797,13 +869,35 @@ export async function getServiceById(
...
+const SERVICE_ID_PATTERN = /^[A-Z0-9]{6,}$/;
+
+export function isValidServiceId(id: string | undefined | null): boolean {
+  if (typeof id !== 'string') {
+    return false;
+  }
+  return SERVICE_ID_PATTERN.test(id.trim());
+}
+
+async function getServicesByIdsBatch(
```
renannoliveira, line 900: **[HIGH] `getServicesByIdsBatch` makes a single non-paginated API call — services beyond the first page are silently dropped.**

The function fetches once and returns `result.services ?? []` with no pagination loop. `getAllServices` wraps its fetch in a `do { … } while (offset < result.total!)` loop — this one doesn't. PagerDuty's default page size is 25; a batch of 50 IDs can match more than 25 services, and anything at positions 26–50 is silently discarded.

Entities backed by those services will always appear as `NotMapped` on the mappings page. Fix: add `limit=50` to the query params to guarantee the full batch fits in one page, or add the same offset/loop used in `getAllServices`.

plugins/backstage-plugin-backend/src/service/router.ts
```diff
@@ -620,6 +624,7 @@ export async function createRouter(
             serviceId: entity.serviceId,
             vendorId: backstageVendorId,
             account: entity.account,
+            cache,
```
renannoliveira, line 627: **[MEDIUM-HIGH] The standalone `POST /services/:serviceId/integration` endpoint at line 1183 omits `cache` from `createServiceIntegration` — the cache is never invalidated after that integration is created.**

This call site and the one at line 727 correctly pass `cache`. The third call site at line 1183 does not. `deleteCache()` is a no-op when `cache` is `undefined` (see `pagerduty.ts` lines 225–233), so neither the per-service nor `ALL_SERVICES_CACHE_KEY` entry is cleared after a standalone integration creation.

Within the 20-minute TTL, `getServiceById` returns the stale cached service (no `backstageIntegration`). A second caller will see no integration and attempt to create a duplicate.

Fix: pass `cache` to `createServiceIntegration` at line 1183, the same way this call site does.

plugins/backstage-plugin-backend/src/apis/pagerduty.ts
```diff
@@ -946,16 +1075,27 @@ export async function getServiceByIntegrationKey(
...
+export async function getAllServices(
+  cache?: CacheService,
+): Promise<PagerDutyService[]> {
+  // Return the cached service list if we already have one; only hit the API
+  // again when there are no cached services.
+  const cached = await readCache<PagerDutyService[]>(
+    cache,
+    ALL_SERVICES_CACHE_KEY,
+  );
+  if (cached && cached.length > 0) {
```
renannoliveira: **[MEDIUM] `cached && cached.length > 0` treats an empty service list as a cache miss — infinite re-fetches for zero-service accounts.**

The cache write at line ~1169 is unconditional: `[]` is written for an empty account. But on the next call `cached` is `[]` (truthy), `cached.length > 0` is `false`, so the function falls through to a fresh full paginated API fetch, writes `[]` again, and the cycle repeats indefinitely.

This is a genuine loop for any staging/test account with zero services, and can also be triggered by a transient API error that returns an empty list.

Fix: `if (cached !== undefined && cached !== null)` (or `if (cached != null)`).

plugins/backstage-plugin/src/components/PagerDutyPage/ConfigurationPage.tsx
```diff
@@ -0,0 +1,115 @@
...
+          }
+        });
+    }
+
+    fetchSetting();
+  }, [pagerDutyApi]);
+
+  const handleChange = (value: StoreSettings) => {
+    setSelectedServiceDependencyStrategy(value);
+    pagerDutyApi.storeSettings([
```
renannoliveira: **[MEDIUM] `pagerDutyApi.storeSettings()` is called without `await` and without `.catch()` — save failures are silently discarded.**

`handleChange` is not `async`. The returned `Promise` is dropped. On a network error or auth failure, `setSelectedServiceDependencyStrategy(value)` has already run (line 64), so the radio button shows the new selection — but the backend was never updated. On next page load, `fetchSetting` re-reads from the API and silently reverts the UI with no error shown to the user.

Fix: make `handleChange` async and `await` the call, or add `.catch(err => { /* show error toast */ })`.

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
```diff
@@ -169,7 +174,35 @@ export function getMappingEntities(store: PagerDutyBackendStore, catalogApi: Cat
...
+      // malformed value (pasted URL, whitespace, typo) can appear. getServicesByIds
+      // sanitizes again before calling PagerDuty, but surface a warning here so the
+      // operator can fix the offending entity annotation. De-duplicate first so each
+      // id is fetched and warned about at most once.
+      const uniqueTrimmedServiceIds = Array.from(
+        new Set(rawPagerDutyServiceIds.map(id => id.trim())),
+      );
+
+      const currentPagePagerDutyServiceIds = uniqueTrimmedServiceIds.filter(id => {
+        if (isValidServiceId(id)) {
```
renannoliveira, line 198: **[MEDIUM] `isValidServiceId` (regex `/^[A-Z0-9]{6,}$/`) is applied to DB-stored mapping IDs — any legacy 5-character PagerDuty service ID stored in the database will be dropped, making that entity appear as `NotMapped` forever.**

`rawPagerDutyServiceIds` (lines 181–186) includes both annotation-sourced and DB-stored `entityMapping.serviceId` values. All are then filtered through `isValidServiceId` here. A stored mapping with a historically-valid 5-char ID (e.g. `P1234`) fails the 6-char minimum, is never fetched, and the fallback lookup at line ~283 returns `undefined`.

Consider applying the regex only to annotation-sourced IDs and fetching DB-stored IDs unconditionally, or surfacing a more visible error in the UI for affected entities.

renannoliveira (commented):
Code review — 7 confirmed findings. See inline comments for details.

plugins/backstage-plugin/src/alpha/pages.tsx
sandornagy517: Not addressed — this finding doesn't apply to the version this PR targets. It assumes `@backstage/frontend-plugin-api` **v0.13.2**, but this PR bumps `backstage.json` to **1.51.2**, and the managed `backstage:^` spec resolves the package to **v0.17.1** (see `yarn.lock`: `@backstage/frontend-plugin-api@backstage:^::backstage=1.51.2&npm=0.17.1`).

In 0.17.1:
- `SubPageBlueprint` **is** exported (it's in the package's public `index.d.ts` export list and ships as its own module), so the `SubPageBlueprint.make(...)` calls resolve correctly — no `TypeError` at import time.
- The developer comment in `pages.tsx` matches the actual implemented behavior: `PageBlueprint`'s factory has a no-`loader` branch that, when sub-pages are attached via the `pages` input, renders them as tabs and adds an index route that `<Navigate>`s to the first sub-page.

So the alpha page does not crash; it renders as intended on 1.51.2. Leaving as-is.

plugins/backstage-plugin/src/alpha/plugin.ts
sandornagy517: Not addressed — same root cause as the `SubPageBlueprint` comment: the analysis is against an older `@backstage/frontend-plugin-api` than the one this PR uses (0.17.1, via Backstage 1.51.2).

In 0.17.1, `PageBlueprint`'s params schema **does** include `title?` and `icon?`, and its factory yields them as `coreExtensionData.title` / `coreExtensionData.icon` — which is exactly what drives the plugin's sidebar nav entry. So consolidating `title`/`icon` onto `PageBlueprint` (instead of a separate `NavItemBlueprint`) is the intended pattern here, and the sidebar entry is present. They are not silently ignored. Leaving as-is.

plugins/backstage-plugin-backend/src/apis/pagerduty.ts
sandornagy517: Fixed in da4ad52. Added `limit=SERVICE_IDS_BATCH_SIZE` (50) to the `id[]` query in `getServicesByIdsBatch`, so the full batch always fits in the single non-paginated request and ids 26+ are no longer dropped. Since the batch size (50) stays under PagerDuty's 100-id ceiling, one page is guaranteed to cover the whole batch — no offset loop needed.

plugins/backstage-plugin-backend/src/service/router.ts
sandornagy517: Fixed in da4ad52. The standalone `POST /services/:serviceId/integration/:vendorId` handler now passes `cache` to `createServiceIntegration` (it was already in scope, destructured from `options` at the top of `createRouter`), matching the other two call sites. The per-service and all-services cache entries are now evicted after a standalone integration is created, so a subsequent read no longer returns a stale service within the TTL.

plugins/backstage-plugin-backend/src/apis/pagerduty.ts
sandornagy517: Fixed in da4ad52. Changed the guard to `if (cached != null)` so an empty cached array (a zero-service account, or a transient empty result) is treated as a valid cache hit rather than a miss, eliminating the repeated full re-fetch.

plugins/backstage-plugin/src/components/PagerDutyPage/ConfigurationPage.tsx
sandornagy517: Fixed in da4ad52. `handleChange` is now `async` and `await`s `storeSettings`. On failure it rolls back the optimistic radio selection to the previous value and surfaces an error via `alertApiRef` (the convention used elsewhere in this plugin, e.g. `TriggerDialog` / `ServiceNotFoundError`), so save failures are no longer silently discarded.

plugins/backstage-plugin-backend/src/controllers/mappings-controller.ts
sandornagy517: Not addressed (intentionally) — the concern is valid in principle, but a fix here is both ineffective at this layer and at odds with existing tested behavior, and there's no affected data in practice:

1. **No data is affected today.** Real PagerDuty service ids are 7 chars (e.g. `P6V1UFE`, `PDLO6HU` in `examples/entities.yaml`). The only ≤5-char ids in this repo are *custom-field* ids (e.g. `PD123`), not service ids.
2. **A controller-only change wouldn't work.** `getServicesByIds` → `getSerivcesByIdsAndAccount` re-applies `isValidServiceId` before calling PagerDuty. That downstream filter exists deliberately: the `id[]` list endpoint returns **400 for the whole batch** if any single value is malformed, so it protects the batch from one poisoned annotation id. Letting DB ids bypass the controller filter alone would just have them re-dropped one layer down.
3. **The 6-char floor is an intentional, tested contract.** `pagerduty.test.ts` asserts `isValidServiceId('PXX') === false` and `isValidServiceId('PXX XXXX') === false`. Relaxing the regex would break that contract.

A correct fix would need an explicit "trusted id" path threaded through the API layer (validate only annotation-sourced ids, pass DB-stored ids unconditionally) plus the corresponding test changes — a larger change than this theoretical edge case warrants. Tracking as possible future hardening rather than addressing in this PR.

renannoliveira (approved):
lgtm

